Issue #53, opened 2020-04-29
Spline knot optimization variable causes issue with VMEC profile initialization


schroeder24:
I am attempting to run STELLOPT with an optimization of temperature spline knots, and am finding problems when I assign a spline knot as an optimization variable. Specifically,  when any knot is labeled as an optimization variable the mass profile is initialized with values 4 orders of magnitude larger than expected. When there is no knot specified as a variable, the mass profile initializes as expected. This behavior seems strange to me.

I attached two input files to this report, the only difference between them is whether or not I include a spline knot as an optimization variable (the input that 'works' initializes VMEC correctly, while the input that is 'broken' leads to VMEC convergence issues because the mass profile is not initialized correctly - I am basing this diagnosis based on the 'am_aux_f' values reported in the temp_input file that is generated at startup). I tried to keep these inputs as simple as possible.

I tested these inputs on Draco, on the Lazerson branch of STELLOPT.

[input.without_knot_var_works.txt](https://github.com/PrincetonUniversity/STELLOPT/files/4554036/input.without_knot_var_works.txt)

[input.with_knot_optvar_broken.txt](https://github.com/PrincetonUniversity/STELLOPT/files/4554037/input.with_knot_optvar_broken.txt)

PR #56, opened 2020-05-07
Merge branch Lazerson into develop

lazersos:
The Lazerson branch contains a number of bugfixes and improvements and is up to date with develop.  The major updates are as follows:

- BEAMS3D various upgrades for generating input for ASCOT4 and ASCOT5.
- Added compilation support and updates for Marconi, Draco, Cobra, HGW_Theory, JFRS1
- Fixes missing SFINCS variables in make_all.inc
- Fixes STELLOPT profile optimization bug #53 
- LIBSTELL should now properly compile in serial
- Dependency files now moved out of Release and Debug, only one file necessary now.
- Various routines removed in accordance with `license` branch.
- Fixed bug in STELLOPT helicity calculations.

Please verify that the code compiles on your machines:

- [x] Macports (GCC, @lazersos )
- [x] Draco (Intel, @lazersos, @jcschmitt )
- [x] Cobra (Intel, @lazersos, @jcschmitt )
- [x] Marconi (Intel, @lazersos )
- [ ] Cori (Intel, @landreman, @zhucaoxiang)
- [x] Sunfire (@zhucaoxiang, @jbreslau, @mdjcole )
- [ ] Wisconsin Machines (@jcschmitt, @benjaminfaber, @aaroncbader)


lazersos:
@zhucaoxiang Running the test with GCC and debug I don't get this error the test goes through just fine.  I'm building and checking on Draco with debug and the Intel compiler now.

lazersos:
@zhucaoxiang I just tested on Draco and found no problem, although I see our debug flags are slightly different.  I'm going to rerun with more flags and check.

lazersos:
@zhucaoxiang I'm able the replicate the issue if I expand my debug flags a bit.  I can also replicate a similar error but running VMEC alone.  I should note this doesn't appear durring a production run.  I suspect this has something to do with VMEC's use of initialized variables which imply an implicit `SAVE`. For example
```
       INTEGER :: test = 1
```
is the same as
```
      INTEGER, SAVE :: test=1
```
Also it should be noted that because of this one should always initialize variables in the main parts of a subroutine, not in the deffinitions.

I'll point this out to Marc but this is clearly a subtle VMEC bug not an issue we should fix in STELLOPT.

zhucaoxiang:
@lazersos That's interesting. I will check your Intel compiler options and see if it works at PPPL. Just in case you don't know, the repo is now fully public. ORNL agrees we use VMEC as part of STELLOPT.

zhucaoxiang:
@lazersos I tested on PPPL clusters. 

- GCC: Both release/debug versions are working
- Intel: The release version seems working, but the debug version has the issues that are reported above.

I would agree to merge, but just keep a record that these issues are actually existing.

Just for confirmation, `NTCC` is no longer required for any part of the code. Right? @lazersos I just removed the dependence in the Makefiles for PPPL.

lazersos:
@zhucaoxiang Good idea.  We should create an issue here to track this bug, and I'll mirror the issue on the ORNL/VMEC Github site as well.

Regarding NTCC.  Yes we should be able to build all codes without needing to link to NTCC anymore.  In fact, the beam deposition routines in BEAMS3D now only need to know where the database files are and no longer rely upon linking to the libraries.  Also we no longer need the preprocessor directive either.

PR #59, opened 2020-05-16
Stellcopt

jbreslau:
A few more updates and fixes to accommodate N. Pomphrey's project needs.


lazersos (changes requested):
We can't do this here.  The `lfreeb` variable only indicates that the code is free boundary.  If I run a free boundary optimization with STELLOPT but not optimizing coils then this will overwrite the name of the file I used to run the code.  This should be done in the subroutine where you generate the mgrid file.

jbreslau:
> We can't do this here. The `lfreeb` variable only indicates that the code is free boundary. If I run a free boundary optimization with STELLOPT but not optimizing coils then this will overwrite the name of the file I used to run the code. This should be done in the subroutine where you generate the mgrid file.

But Neil specifically requested a change to the name appearing in the generated input file.  Would it solve the problem if I changed 'lfreeb' to 'lcoil_geom' in the conditional that changes the mgrid file name?  That would prevent it from overwriting your file name in non-coil free boundary cases.

lazersos:
> > We can't do this here. The `lfreeb` variable only indicates that the code is free boundary. If I run a free boundary optimization with STELLOPT but not optimizing coils then this will overwrite the name of the file I used to run the code. This should be done in the subroutine where you generate the mgrid file.
> 
> But Neil specifically requested a change to the name appearing in the generated input file. Would it solve the problem if I changed 'lfreeb' to 'lcoil_geom' in the conditional that changes the mgrid file name? That would prevent it from overwriting your file name in non-coil free boundary cases.

Yes I think that's a better choice.

jbreslau:
The requested change has been made.

lazersos (approved):

PR #64, opened 2020-06-12
Lazerson

lazersos:
This pull request has a few key updates.

- Parallelization in STELLOPT is now more flexible allowing for NOPTIMZIERS to be set more or less arbitrarily (can split nodes, or join them).
- BEASM3D/ASCOT5 updated to support v5.3 (dummy MHD and BOOZER datasets added).
- Suzuki beam stopping model now added to BEAMS3D so PREACT/ADAS no longer required.
- BEAM3D now outputs user controllable 5D distribution function.
- New tree based wall model module implemented.
- Added new dp/ds target in STELLOPT
- Few bug fixes for VMEC and STELLOPT


zhucaoxiang:
@lazersos We should probably merge it at some point. It not a good idea to keep too many things in one pull request.

PR #66, opened 2020-06-18
Added polygonal keepout zones as an optional coil chisq target

jbreslau:
New coil penalty function added at N. Pomphrey's request.


lazersos (approved):
Let's get this merged in.

PR #69, opened 2020-06-24
Improve pySTEL to automatically check dll names

zhucaoxiang:
I have inserted a function to automatically assign the correct strings for ctypes functions and variables. This is a quick fix for not changing the source code and style.

Personally, I don't think putting `import ctypes`, `libstell`, etc. in the namespace of the file is an issue. It will avoid a lot of reduplications. But this requires a lot of changes. 

I am not quite sure if this will affect the usage of pySTEL on Windows.


zhucaoxiang:
I accidentally polluted this branch with CI implementations. I will close it for now and re-open it with clean commits.

lazersos:
I'm taking the today and tomorrow off but I'll test this on the following as I have time:

- [x] Macports OSX
- [x] Draco (MPCDF)
- [x] Cobra (MPCDF)
- [x] Marconi (Cineca)

lazersos:
So thing look good for Intel.  One problem I have with GNU is that at least for Macports the namespaces have three leading underscores but the ctypes interface only wants two in the deffinition of s1.  I've tried `-fno-leading-underscore` but then linking fails (presumably because all other libraries were built without it). I'll explore this in more detail over the coming days.

zhucaoxiang:
@lazersos The user can still insert their customized strings in `__init__.py` after the line `s1, s2, s3 = get_s123()` to override the automatic values.

lazersos:
The behavior is really weird on MacOS with GNU compilers built by macports.  The lines read like `s1='___'`, `s2='MOD'`, and `s3=''` (which is correct).  However, the ctypes interface only works if `s1='__'` (three vs two underscores).  I've added a line which checks for three underscores and defaults to two for now.  Intel and GNU and Intel compilers are working.

Note that I tried added `-fno-leading-underscore` but then I get a bunch of errors from EzCDF.  Anyway, this should be fine as GCC and Intel are the two major compilers people use.  I'd be interested to know if this is a MacOS GNU issue or a GNU issue in gener

zhucaoxiang:
@lazersos I was using the same solution as you did. Just manually correct it for now.

PR #76, opened 2020-07-15
Develop sfincs 2020

jcschmitt:
Changes/Updates to the stellopt/sfincs coupling to handle shared memory groups.

stellopt_sfincs:  MPI_COMM_SFINCS is not split from MPI_COMM_MYWORLD  instead of  MPI_COMM_WORLD

A new variable has been added to stellopt_runtime called 'id_tag'.  This variable holds the original input file name. Other variables that contain this name may be modified during runtime, but this variable should not be changed after the 1st assignment.  

SFINCS (during the vboot iteration) copies portion of the namelist form the original input file on each iteration. (variable 'id_tag' holds this string).  See stellopt_vboot.f90 and stellopt_sfincs.f90 

Some documentation was updated. (stellopt_fcn, stellopt_init)


jcschmitt:
Has anyone had a chance to look at the modifications to see if they are appropriate?

PR #80, opened 2020-07-29
Feature/regcoil updates 2019-2020

jcschmitt:
Increased the number of variables/targets/metrics that STELLOPT could use from REGCOIL calculations.

Also includes a bug fix.


jcschmitt:
Out of date request. Other development activities have made this merge unnecessary.

PR #90, opened 2020-12-29
Hotfix nyquist modes in stellopt_kink

zhucaoxiang:
@lazersos modified `stellopt_kink` to use the Nyquist mode (check #42 ), but a missing line will cause a problem that `MLMNV` on different CPUs are different and thus a divided_by_zero error will occur.


lazersos:
Looks good to me....good catch.

lazersos (approved):
Looks good.

PR #93, opened 2021-01-08
Bug fix for PEST coordinate calculations.  

jcschmitt:
This bug fixes a problem for inversions near the poloidal angle = 0,2*pi boundary.


lazersos (approved):
Looks OK to me.

PR #99, opened 2021-02-23
Hotfix BNORM interface to adopt nyquist mode changes in VMEC

zhucaoxiang:
As mentioned in #42 , the default value for `lnyquist` is now `true` and it has been removed from the input namelist. A bunch of codes that are interfacing VMEC now need to adopt the changes. Here is the fix to BNORM. (PPPL cluster module has been updated).

Here is a simple comparison using the half-Tesla NCSX LI383 configuration. It looks like the updates fix the issue for now.
- lnyquist = .false.
image
- lnyquist = .true. (inconsistent)
image
- lnyquist = .false. with BNORM interface updated (consistent)
image


lazersos:
Shouldn't the third plot be labeled `lnyquist=.true.` if the problem is fixed?

zhucaoxiang:
> Shouldn't the third plot be labeled `lnyquist=.true.` if the problem is fixed?

Yes, it is with `lnyquist=.true.`. I was using 'on/off' when generating the figures, which stands for "T/F".

lazersos (approved):
Looks fine to me.

jcschmitt:
I think terpsichore was turned on in this commit - I guess this will test the rest of the logic flags, but I don't think most people have terps by default.

PR #104, opened 2021-03-07
STELLOPT: Fixes issue with radial weighting of helicity

### LINKED ISSUE #95: Weighting of Fourier modes in quasisymmetry target
When targeting quasisymmetry in stellopt, `chisq_helicity.f90` applies a surface-dependent weight to each Fourier mode in the Boozer spectrum. At first glance, it appears that this weight also depends on the poloidal mode number m:

https://github.com/PrincetonUniversity/STELLOPT/blob/38c832062fb298165b07f96a0627f1f0f1d16c01/STELLOPTV2/Sources/Chisq/chisq_helicity.f90#L97-L105

However, this bit of code is not in a loop over Fourier modes, so m in these lines ends up being the last value of m from a loop above (lines 61-92), not the actual value of m for each given mode. In contrast, when `TARGET_HELICITY_OLD` is used instead of `TARGET_HELICITY`, so `chisq_helicity_ornl.f90` is run, it appears that the weight is actually m-dependent:

https://github.com/PrincetonUniversity/STELLOPT/blob/38c832062fb298165b07f96a0627f1f0f1d16c01/STELLOPTV2/Sources/Chisq/chisq_helicity_ornl.f90#L74-L79

That logic in `chisq_helicity.f90` is misleading, so it should be corrected or removed.

It also seems suspicious that in `chisq_helicity.f90` the radius-dependent weight is applied only when sigma is >0, whereas in `chisq_helicity_ornl.f90` the radius-dependent weight is applied only when sigma is <0.

lazersos:
Bugfix for the helicity chisq metric.

### DIFF
--- a/STELLOPTV2/Sources/Chisq/chisq_helicity.f90
+++ b/STELLOPTV2/Sources/Chisq/chisq_helicity.f90
@@ -35,7 +35,8 @@
 !-----------------------------------------------------------------------
       LOGICAL :: lsym
       INTEGER :: dex, ik, mn, n, m, k_heli, l_heli, num0, n1, n2, i_save
-      REAL(rprec) :: bnorm, bmax, bmn, rad_sigma, sj, val
+      REAL(rprec) :: bnorm, bmax, bmn, sj, val
+      REAL(rprec), DIMENSION(:), ALLOCATABLE :: rad_sigma
 !----------------------------------------------------------------------
 !     BEGIN SUBROUTINE
 !----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


zhucaoxiang:
@lazersos Thanks for the fix! After changing `rad_sigma` from a scalar to an array, the following line might cause trouble if `rad_sigma` and `sigmas(num0:mtargets) ` are not in the same dimension. It looks good to me (by eye). But you can double-check it or run a test case.

```
sigmas(num0:mtargets)  = ABS(sigma(ik))*rad_sigma*sigmas(num0:mtargets)
```

lazersos:
@zhucaoxiang I believe the way the code is organized the two are always the same.  As `num0` get set to `mtargets` and `mtargets` is always incremented by `mnboz_b`.

zhucaoxiang (approved):

PR #125, opened 2021-05-16
Fix errant points in U_lines from BEAMS3D

landreman:
Presently, the U_lines output from BEAMS3D has errant points. For example, here are some trajectories from a perfectly axisymmetric VMEC configuration, computed using `develop` compiled just now:
image
This problem makes it hard to display and understand trajectories in flux-coordinate space. The problem seems to be that U_lines is obtained in `out_beams3d_nag.f90` by direct interpolation of U_ARR, which has a discontinuity on the line to the right of the magnetic axis, where u changes from 2pi to 0. This problem is fixed in this pull request. Now instead of performing interpolation on u, interpolation is performed on x and y, defined by x= sqrt(s) * cos(u) and y = sqrt(s) * sin(u) (so x and y are smooth functions of position), and U_lines is determined by atan2(y, x). Here is the output for the same axisymmetric trajectories after this fix:
image
It appears that interpolation directly on U is also performed in `beams3d_physics_mod.f90` around line 1362, and this may need to be fixed as well.


BEAMS3D/Sources/beams3d_init.f90
```diff
@@ -432,10 +438,20 @@ SUBROUTINE beams3d_init
...
+         X_spl%x3 = zaxis
+         Y_spl%isHermite = 1
+         Y_spl%x1 = raxis
+         Y_spl%x2 = phiaxis
+         Y_spl%x3 = zaxis
          POT_spl%isHermite = 1
          POT_spl%x1 = raxis
          POT_spl%x2 = phiaxis
          POT_spl%x3 = zaxis
+         X_ARR = sqrt(S_ARR) * cos(U_ARR)
```
lazersos, line 453: Do we need the sqrt(S_ARR) factor?  Or can we just use `cos(U_ARR)` and `sin(U_ARR)`?

lazersos (approved):
Overall looks fine to merge into develop.

lazersos:
I'm going to merge this in.  In the `Lazerson` branch I've merged these changes as well but then further modified the code to remove the `U_spl`, `U4D`, and `U_ARR` arrays completely.  The code still outputs `U_ARR` but constructs it from `ATAN2(Y_ARR,X_ARR)`.

PR #139, opened 2021-09-21
added make_arch_linux.inc, removed unneccesary flags from make_ubuntu

smiet:
The make_ubuntu.inc included `-lblacs` etc flags that are unneccesary on the latest version of ubuntu, as blacs is currently included in the scalapack libraries. 

Tested that this .inc file compiles sucessfully on ubuntu 20.4.3 LTS with the libraries specified [here](https://princetonuniversity.github.io/STELLOPT/STELLOPT%20Compilation%20Ubuntu) all installed

Also added the arch_linux.inc file with changed links, and the `-fallow-argument-mismatch` flag necessary for compilation with gcc>9

### DIFF
--- a/SHARE/make_arch_linux.inc
+++ b/SHARE/make_arch_linux.inc
@@ -1,10 +1,10 @@
 #######################################################################
 #            Define Basic Utilities
 #######################################################################
-  SHELL = /bin/sh
+  SHELL = /bin/zsh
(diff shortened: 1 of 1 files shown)


lazersos (approved):
Looks simple and straightforward.  Shouldn't break the code for anyone.

PR #143, opened 2021-11-05
EPS_EFF label in STELLOPT updated to EPS_EFF^(3/2)

landreman:
When NEO is used via STELLOPT, STELLOPT presently prints values to the screen under the label `EPS_EFF` that are actually epsilon_eff^(3/2). This pull request has a few changes so users do not misinterpret these numbers as epsilon_eff^1. The label in STELLOPT is changed to `EPS_EFF^(3/2)`, and the variable name `eff_ripple` is updated to `eps_eff32`.

The numbers printed to stdout and saved in the `neo_out` and `neolog` files come from the variable `epstot` in NEO. To see that `epstot` in NEO is actually epsilon_eff^(3/2) rather than epsilon_eff^1, here's where epstot is computed:
https://github.com/PrincetonUniversity/STELLOPT/blob/develop/NEO/Sources/flint_bo.f90#L426-L429
where `coeps` has the factors pi R^2 / (8 sqrt(2)) from  eq (29) in Nemov, PoP 6, 4622 (1999); doi: 10.1063/1.873749:
https://github.com/PrincetonUniversity/STELLOPT/blob/develop/NEO/Sources/flint_bo.f90#L135
Comparing to Nemov's (29), epstot must be eps_eff^(3/2).

This definition is also documented in NEO:

https://github.com/PrincetonUniversity/STELLOPT/blob/develop/NEO/Sources/flint_bo.f90#L17

### DIFF
--- a/STELLOPTV2/Sources/Chisq/chisq_neo.f90
+++ b/STELLOPTV2/Sources/Chisq/chisq_neo.f90
@@ -12,7 +12,7 @@
       USE stellopt_runtime
       USE stellopt_targets
       USE stellopt_vars, ONLY: equil_type
-      USE equil_vals, ONLY: eff_ripple
+      USE equil_vals, ONLY: eps_eff32
 !DEC$ IF DEFINED (NEO_OPT)
       USE neo_input_mod, ONLY: read_neoin_input, write_neoin_namelist
 !DEC$ ENDIF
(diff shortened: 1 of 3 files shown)


lazersos (approved):
Good idea, thanks for clarifying this in the code.

PR #145, opened 2022-03-03
Kulla fidasim

lazersos:
Upgrades and improvements

- First pass at an interface to FIDASIM
- Improved interface to EQDSK G-files


lazersos:
OK I think this version is ready for testing now.  @kudav @alexleviness if you could verify a case you've done before then we can go ahead and merge it in.  I'll do the same.

alexleviness (approved):
I verified, looks fine to me!

PR #155, opened 2022-04-29
Kulla fidasim

kudav:
Fixed and partially verified version of the FIDASIM interface with correct distribution output (see below).

grafik


lazersos (approved):
Looks fine to me.

PR #161, opened 2022-09-27
Kulla fidasim

kudav:
Production version of the FIDASIM interface in BEAMS3D with optional distribution function on a cylindrical grid (-fidasim2 flag).
Additionally includes improved coulomb logarithm formulations.
Should there be the option to pick a formulation on a per-run basis?

Also implements the EQDSK interface from BEAMS3D in the FIELDLINES code.


lazersos (approved):
Overall looks good.  I approve this commit.

PR #173, opened 2023-04-28
LIBSTELL: cyl2flx fix initial guess

jonathanschilling:
`cyl2flx` (in in there `newt2d`) fails me for a FIELDLINES W7-X case if this hotfix is not in place.


lazersos (approved):
This isn't terrible but I really would like to avoid if statements inside loops. This could deal with the reflection of the 
```Fortran
sflux = MAX(xc_opt(1),zero)*SIGN(one,xc_opt(1))
```
but I can't come up with a more efficient way to do the second part.

PR #175, opened 2023-06-02
FIELDLINES integration into BEAMS3D

lazersos:
I've added the ability to base a BEAMS3D run on a run of the FIELDLINES code.

- Magnetic field is read from the FIELDLINES file (allows for islands and stochastic regions)
- Flux grid is contructed from Poincaré trace treating the first fieldine as the magnetic axis (like that produced by a -full run of FIELDLINES).
- Note FIELDLINES should be run with NPOINC >= NPHI, no interpolation for now.
- 'Holes' in rho can still show up if doing a -full run for out of phase islands.  Best to base on a manual specification of field line starting points.


kudav:
Did not work with a -restart run for me on raven. All nodes except the first gave an invalid MPI_Win error. Deposition did work with EQDSK but required a bit of care.

lazersos:
Various improvements have now been added.  FIELDLINES now has a `-gridgen` option which attempts to more robustly choose field line starting locations.  The first element of the `_START` arrays are still used as a guess for the magnetic axis.

lazersos:
I'm going to merge this in as soon as the tests complete.

PR #176, opened 2023-06-09
fix: build error on ubuntu

dboeckenhoff:
On ubuntu, the build failed. Credit to @lazersos


kudav (approved):
The suggested change seems a bit more strict, so it would be good to use it and it compiles for me without issue. Long term we should move towards completely removing the necessity for this flag.

lazersos:
Fixed.

pponsv:
Hi, this change breaks compilation for me (compiling via ./build_all). Reverting to `-fallow-argument-mismatch` fixes it.

An example of the errors I get:
```
ezcdf_GenGet.c:163:45:

......
Error: Type mismatch between actual argument at (1) and actual argument at (2) (REAL(8)/COMPLEX(8)).
```

PR #188, opened 2023-09-06
Feature/beams3 d beamdensity

lazersos:
Adds a beam density calculation when neutral beam deposition is used.


kudav:
I changed the MPI communicator in the Barrier to fix an error. It works now locally and also on raven with 4 nodes.

BEAMS3D/Sources/beams3d_beam_density.f90
```diff
@@ -0,0 +1,121 @@
...
+
+      ! Barrier so  calculation is done
+#if defined(MPI_OPT)
+      CALL MPI_BARRIER(MPI_COMM_SHARMEM,ierr_mpi)
+#endif
+      ! Now Divide by the grid volumes
+      CALL MPI_CALC_MYRANGE(MPI_COMM_SHARMEM, 1, (nr-1)*(nphi-1)*(nz-1), mystart, myend)
+      DO s = mystart,myend
+         i = MOD(s-1,nr-1)+1
+         j = MOD(s-1,(nr-1)*(nphi-1))
```
kudav, line 100: We should ensure that the phi grid isnt too coarse to give an artificially low density, i.e. when dphi*raxis(nr) is larger than the beam diameter, the volume average is lowered.

BEAMS3D/Sources/beams3d_beam_density.f90
```diff
@@ -0,0 +1,121 @@
...
+         pl = MODULO(pl,phimax)
+         LGRID = .FALSE.
+         DO s = 1, nl
+            ! Find gridpoint but use half grid
+            i = MIN(MAX(COUNT(raxis-hr2 < rl(s)),1),nr)
+            j = MIN(MAX(COUNT(phiaxis-hp2 < pl(s)),1),nphi)
+            k = MIN(MAX(COUNT(zaxis-hz2 < zl(s)),1),nz)
+            LGRID(m,i,j,k) = .true.
+         END DO
+         WHERE(lgrid) BEAM_DENSITY = BEAM_DENSITY + denbeam
```
kudav: In my understanding, point 0 is the position on the acceleration grid, and point 1 the ionization location. Thus, we need to reduce denbeam by at least a factor of nl to make the calculation realistic, but should really try and figure out the intersection length with the grid cell.

kudav (changes requested):

BEAMS3D/Sources/beams3d_beam_density.f90
lazersos: A error message is now printed if the choice of grid is too low.

BEAMS3D/Sources/beams3d_beam_density.f90
lazersos: Added an warning message if this is an issue.

BEAMS3D/Sources/beams3d_beam_density.f90
lazersos: Fixed in latest push.

PR #192, opened 2023-09-07
Feature/BEAMS3D interaction improvements

kudav:
This collects non-feature changes to the BEAMS3D code from my old branches:

- Added 2 specific errors for Coils and -beamlet flag
- Improved output of the terminal messages
- Distribution axes are now written to output file
- Logic for handling input flags moved to interface_mod


BEAMS3D/Sources/beams3d_input_mod.f90
lazersos, line 1: Why was this added?

BEAMS3D/Sources/beams3d_interface_mod.f90
lazersos, line 1: I don't like the idea that one flag overrides another.

BEAMS3D/Sources/beams3d_runtime.f90
lazersos, line 1: Don't think we need these.  The coil one we really don't need.  I remvoed it.

lazersos (changes requested):
Few things need to be removed or explained.

BEAMS3D/Sources/beams3d_input_mod.f90
kudav: To make it transparent that setting dex_beams requires use of -beamlet when not restarting particles or giving inital conditions. Looking at this again it should also check for beam definition in the namelist.

BEAMS3D/Sources/beams3d_interface_mod.f90
kudav: Then we need to either throw an error or inform about what the code will do. There are and will be input flags and/or namelist inputs which cant be combined or require special treatment.

BEAMS3D/Sources/beams3d_runtime.f90
kudav: The idea was to have an error when there is a problem with the coil file, but the implementation wasn't good.

BEAMS3D/Sources/beams3d_input_mod.f90
lazersos: I'm not sure about the logic there.  This could break other things.  Let's just not add this.

BEAMS3D/Sources/beams3d_interface_mod.f90
lazersos: You may want to do both beam and restart in the same run in the future and handle it.  Remove this.

BEAMS3D/Sources/beams3d_runtime.f90
lazersos: Well COIL_ERR was throwing a coil error when there was a namelist error not a coil error. Just remove these.

BEAMS3D/Sources/beams3d_init.f90
kudav, line 1: Were all of these changes on purpose? This is essentially an old (outdated?) version of the function again.

lazersos (approved):
Looks good.

PR #194, opened 2023-09-07
Feature/BEAMS3D max S

kudav:
This implements functionality in BEAMS3D to be able to specify profiles for S>1. Each profile (ne, te, ti, ni, zeff, pot) can have different maximum S values. The global maximum S value is set by the new S_MAX parameter in the namelist and is defaulted to 1. The rest of the maximum values are determined from the profiles given in the namelist. This information is also printed in the terminal output.
Note that, as there is no nni parameter in BEAMS3D, there is also no s_max_ni. Instead, the length of the ion density profiles is the same as the zeff profile, so s_max_zeff is used instead.


lazersos (approved):
Made a couple of improvements here and there.

PR #195, opened 2023-09-07
Bugfix/BEAMS3D fidasim

kudav:
This fixes and improves various things for the FIDASIM interface:

- Profiles are now also written to the equilibrium
- E-Field and Et is correctly calculated
- the -fidasim2 flag (for native cylindrical distribution) also activates the normal interface
- Energy bins are calculated for largest mass
- MPI and bug fixes


lazersos (approved):
The changes seem OK.  I need to run the benchmarks on this repo to verify.  We should also think of adding a regression test to the benchmarks folder.

PR #198, opened 2023-09-13
Bugfix/beams3 d fidasim

kudav:
- Made the fidasim_inputs_b3d namelist search case-insensitive
- Increased maximum channel number to 512
- Removed unecessary check for input namelist, as this is always required from beams3d


lazersos (approved):
I approve this!

PR #207, opened 2023-09-25
Bugfix/beams3d fidasim

kudav:
- Deleted previous fidasim interface function
- Added back impurity_charge to namelist output
- User specified equilibrium_file, distribution_file, neutrals_file and geometry_file are no longer overwritten


lazersos (approved):
Looks good to me.

PR #210, opened 2023-11-13
BEAMS3D: Pointers nullified so ASSOCIATED statements work.

### LINKED ISSUE #200: -fidasim_cyl flag crashes for W7-X restart run after depo
FIDASIM for a W7-X case crashes when using the -fidasim_cyl flag.  The -fidasim flag works fine.

The issue can be replicated by doing a restart from from a deposition run for W7-X.

lazersos:
This branch fixes an issue in BEASM3D due to the way Fortran handles pointers.  We use pointers to arrays to do allocation of shared memory through MPI.  However, Fortran does not automatically nullify pointers.  So when we call `ASSOCIATED` on the pointers, it returns true even if the pointer wants associated a shared memory. This chnage adds a routine where all pointers in the code can be nullified at the very beginning.  Should fix #200 and another undocumented bug regarding the output of the beam density.

### DIFF
--- a/BEAMS3D/BEAMS3D.dep
+++ b/BEAMS3D/BEAMS3D.dep
@@ -11,6 +11,9 @@ beams3d_beam_density.o: \
 
 beams3d_interface_mod.o: \
       beams3d_input_mod.o \
+      beams3d_runtime.o \
+      beams3d_grid.o \
+      beams3d_lines.o \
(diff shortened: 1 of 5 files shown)


kudav (approved):
I added some missing pointers, looks good to me otherwise

PR #211, opened 2023-11-21
BEAMS3D: fixed pointers in FIDASIM namelist part

kudav:
This removes pointer variables from the BEAMS3D namelist (for FIDASIM) which should never have been there in the first place.


lazersos (approved):
Approved.

Issue #140, opened 2021-10-16
Type mismatch errors compiling with GCC 10 via MacPorts


ChristopherMayes:
Using macports with gcc10

```
sudo port install openmpi-gcc10 
sudo port select --set mpi openmpi-gcc10-fortran
```

`./build_all` results in many errors like:
```
Error: Type mismatch between actual argument at (1) and actual argument at (2) (REAL(4)/COMPLEX(4)).
ezcdf_GenGet.f90:960:40:

  960 |   status = nf_get_var_double(ncid,varid,varval)
      |                                        1
......
  982 |   status = nf_get_var_double(ncid,varid,varval)
```

Note that these do not show up with gcc9.

lazersos:
With GCC10 there is [stricter type checking](https://gcc.gnu.org/gcc-10/changes.html).  From this [Stackoverflow thread](https://stackoverflow.com/questions/63892055/fortran-error-type-mismatch-between-two-unrelated-subroutine-calls) we can try adding `-fallow-argument-mismatch` to the include `FLAGS_R` and `FLAGS_D` lines in `SHARE/make_macports.inc`.  Let us know if this works and we can update the scripts moving forward.

lazersos:
I can confirm that using the `-fallow-argument-mismatch` with GCC10 allows the code to compile.

PR #215, opened 2023-12-29
Arch variable and gfortran>9

krystophny:
Minor modification of `make_ubuntu.inc` so that

1. Architectures besides `x86_64` (e.g. `aarch64`) work,
2. GNU Fortran > 9 version doesn't throw argument mismatch errors (see also #139, #140)

Tested on Docker image on an M1 Mac. The editor also removed some unnecessary whitespaces.


lazersos (approved):

krystophny:
Thanks, Sam!

PR #218, opened 2024-01-13
Bugfix/debug fixes

### LINKED ISSUE #193: BEAMS3D: strange behavior when compiling in debug
When having compiled in debug, e.g with 'make clean_debug', BEAMS3D shows weird behavior. Particularly, GetBCyl in init_vmec does not produce sensible values as it does when compiled with release. One effect is that S_ARR has values of 1.01 inside the Plasma (S<1), but I've also seen it being equal to 4, which is sflx*sflx*scaleup*scaleup=1*1*2*2.
This did not appear to change when going back on the develop branch a couple of months, so does not seem to be a recent issue.

lazersos:
These bugfixes essentially fix issues associated with runtime errors when compiling with the Intel compilers.  The fixes are mostly related to BEASM3D but will proably propagate to FIELDLINES and STELLOPT. Essentially these fixes help patch runtime errors which Intel throws when `-check all` is used.  Most of the errors had to do with divide by zero type issues where the result of the divide by zero was throw away anyway.  There is a change in wall_mod which could slow down the generation of accelerated walls but otherwise most of the fixes speed up the code if anything.  A bug in `beams3d_diagnostic.f90` related to using `REAL` instead of `REAL(rprec)` was also addressed. This should close #193.


LIBSTELL/Sources/Modules/wall_mod.f90
kudav, line 1: Does the intersection calculation produce correct values for these large (almost infinite) dxyz values? I think we may need to handle these cases specifically, as I did here 2d62e6ed6cb002f1e71cd915865bc4f89898d2b4 but perhaps a bit more elegantly.

LIBSTELL/Sources/Modules/ez_hdf5.f90
kudav, line 1: Is this to support very long strings?

kudav (commented):
When trying to run a simulation, I get an error on line 351 of beams3d_physics mod (floating point invalid operation). I addressed a similar issue in fdfde239901b92b3520dd18930e4fc3d6ae15e97 by enforcing s>=0 in the orbit following, but not sure if this is what we want.

kudav (approved):

PR #219, opened 2024-01-17
LIBSTELL: large array fix for MPI

kudav:
As mentioned in #200, the implementation in LIBSTELL currently causes overflows for arrays exceeding the standard integer limit. This fix introduces a typecast such that the window size is calculated without overflow.


lazersos:
I've got a better way.

PR #228, opened 2024-02-20
Feature/beams3 d restart grid

kudav:
This implements the "-restart_grid filename" flag, which reads the (possibly modified) B_R, B_PHI, B_Z, S_ARR and U_ARR fields from a previous BEAMS3D run to conduct the simulation on. Kinetic profiles are supplied by the input namelist as usual, and interpolated according to the given S_ARR as with the other equilibrium options.


lazersos:
I'm going to say we should wait a few days on this.  What I'd like to do is implement a continue functionality which does two things.  First it'll initialize the background grids from a previous BEAMS3D run.  Second, it'll intialize the particles and distributions from a previous run, to continue the run.  The results are then written back into the continue file.  In this way a run can be continued.  To do this I'll need to implement a `beams3d_continue_particles`, `beams3d_continue_wall`, and probably change `beams3d_restart_grid` to `beams3d_continue_grid`.  Then the restarting of a grid just becomes a suboption of the others.

kudav:
I need to use this branch together with the new RHS for AUG simulations. Should I change the function naming and the flag name to -continue, and we can add the other features to that once we get to it? Or should I just merge the branches locally for now?

lazersos:
Go ahead and change the function names and add flag -continue.

PR #232, opened 2024-04-27
BEAMS3D: Minor bugfix and improvements for DISTNORM

lazersos:
Minor improvements to how distnorm works.

### DIFF
--- a/BEAMS3D/Sources/beams3d_diagnostics.f90
+++ b/BEAMS3D/Sources/beams3d_diagnostics.f90
@@ -171,18 +171,12 @@
             dense_prof(:,k)  =  dense_prof(:,k)/vp_temp
             j_prof(:,k)      =      j_prof(:,k)/vp_temp ! [A/m^2]
          END DO
-         ! Normalize to velocity space volume element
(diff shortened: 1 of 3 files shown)


kudav:
grafik
I tested a W7X case with no collisions (12 Particles, same initial conditions, no losses) and there does seem to be a quite subtle difference in the distribution. I wonder if it is because of the change in the toroidal indexing? 

The plots are generated with the 'denf2d' flag of the plot_beams function, which calculates the plotted data as follows:
dist = squeeze(trapz(beam_data.paxis,trapz(beam_data.Vaxis,trapz(beam_data.Waxis,sum(beam_data.dist_prof,1),6),5),4));
I've confirmed that the axis arrays are the same, but maybe the definition was not consistent before these fixes?

lazersos:
Exactly.  This was a bug was with respect to how the toroidal angle was being calcualted in the distribution normalization.  Shouldn't affect NFP=1 configurations but does affect stellarators.  Thankfully this only affected the 5D distribution function which wasn't really used in any published results.

kudav (approved):
Looks good

PR #233, opened 2024-05-12
[DIAGNO] fix db_midpoint

jonathanschilling:
errornously stored result in `dflux_midpoint` as return variable name


lazersos (approved):
Good catch. Personally had always used the simpson method.

PR #234, opened 2024-05-12
[DIAGNO] add an option to write a more accurate output file for the bfield_points diagnostic

jonathanschilling:
With this patch, the user can use a `DIAGNO_IN` namelist flag `lbpoints_accurate_output` to activate a different output file format for the `diagno_btest.<runId>` output file, which has:
1. the number of points that were written in the first line (for easier parsing, similar to the fluxloop and segmented Rogowski output files) and
2. more digits of accuracy for each of the outputs (the output format was copied from the corresponding writing routine for the segmented Rogowski sensors)

By default, this option is disabled in order to retain backwards compatibility.


lazersos (approved):
Looks fine to me.

PR #236, opened 2024-05-13
Feature/b3d hamiltonian rs develop

kudav:
This implements a new right hand side for the guiding center equation of motion (fgc_eom), which can be swapped out for fgc_nag in the wrapper fgc_lsode (or the rhk68 functions). The formulation is derived in Littlejohn 1983 (J Plasma Phys.) and conserves energy much more accurately than the previous formulation.


lazersos (approved):
This looks good to me.

PR #237, opened 2024-05-13
[DIAGNO] add an option to write a more accurate output file for the afield_points diagnostic

jonathanschilling:
[Similar as #234, but this time for the new `afield_points` diagnostic. Started working on that separately and wanted to wait for feedback on 234, thus only submitting now.]

With this patch, the user can use a `DIAGNO_IN` namelist flag `lapoints_accurate_output` to activate a different output file format for the `diagno_atest.<runId>` output file, which has:

* the number of points that were written in the first line (for easier parsing, similar to the fluxloop and segmented Rogowski output files) and
* more digits of accuracy for each of the outputs (the output format was copied from the corresponding writing routine for the segmented Rogowski sensors)

By default, this option is disabled in order to retain backwards compatibility.

Also, added writing of the flags `lapoints_accurate_output` and `lbpoints_accurate_output` to `write_diagno_input`.

### DIFF
--- a/DIAGNO/DIAGNO.dep
+++ b/DIAGNO/DIAGNO.dep
@@ -44,6 +44,14 @@ diagno_init_spec.o : \
       diagno_runtime.o
       
 
+diagno_afield.o : \
+      ../../LIBSTELL/$(LOCTYPE)/stel_kinds.o \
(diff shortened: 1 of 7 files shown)


lazersos (approved):
Looks fine to me.

PR #238, opened 2024-05-17
STELLOPT: Issue where iflag/=0 was not being caught by subcodes.

lazersos:
This fixes an error where iflag was being overwritten by ier_paraexe.  By placing the `iflag = ier_paraexe` inside a IF/ENDIF pair the issue is fixed.


leebr48 (approved):
Changes look good.

PR #240, opened 2024-06-03
Use a more direct (i.e. no loops) implementation of `mpi_calc_myrange`

jonathanschilling:
This one also checks that no errornous `mystart`, `myend` are created if more ranks are tried to be assigned than work items (e.g. loop iterations, field lines to be traced, ...) are available.


lazersos:
This is a fundamental and substantial change to a piece of code used by many routines. I'm guessing this arrises because you're using more processors than you have work.  Not sure why this would arrise.  Please raise an issue with an input file which can demonstrate the bug.

lazersos:
I'm going to close this pull request becasue the behavior of the proposed code is poor.  As this subroutine shows up in many places it must ALWAY run. Meaning no STOP statements. If there isn't enough work then `mystart` should simply be `> myend` so any `FOR i = mystart,myend` will simply be skipped over and the work still performed using the available processors.

PR #241, opened 2024-06-03
[FIELDLINES] allow start from plasma boundary given in VMEC `INDATA`, not only `wout`

jonathanschilling:
These changes allow to provide the command-line argument `-edge` to FIELDLINES not only in combination with tracing over a VMEC equilibrium, but also in vacuum (`-vac`).
The plasma boundary geometry is then read from the VMEC `INDATA` namelist using `RBC` and `ZBS` (and in case of `lasym`, also `RBS` and `ZBC`). Furthermore, `nfp`, `mpol` and `ntor` are needed.

Additional input parameters (in the `FIELDLINES_INPUT` namelist) named `nstart_pol` and `nstart_tor` are provided to specify how many grid points over the plasma boundary are to be used in each of the tangential directions. They default to 0, in which case (as if they are not specified in the `FIELDLINES_INPUT` namelist) operation defaults to `int(floor(sqrt(maxlines)))` per direction, as previously implemented.


jonathanschilling:
Note: Assuming one has an existing `wout` file, but only wants to use the plasma boundary geometry from there, one can copy that plasma boundary geometry into a new VMEC `indata` namelist using the `wout2indata` tool from #245.

jons-pf:
@lazersos FYI, I think this should now be compatible again with the recently merged py_stel update.

LIBSTELL/Sources/Modules/fieldlines_input_mod.f90
```diff
@@ -176,6 +185,8 @@ SUBROUTINE write_fieldlines_namelist(iunit_out, istat)
       WRITE(iunit_out,outflt) 'FOLLOW_TOL',follow_tol
       WRITE(iunit_out,outint) 'NPOINC',npoinc
       WRITE(iunit_out,outflt) 'MU',mu
+      WRITE(iunit_out,outint) 'NSTART_POL',nstart_pol
```
lazersos, line 188: These should only be written out if they're >0.

lazersos (changes requested):
The modifications in FIELDLINES_INIT_MGRID and COIL (All the BCASTS) shouldn't be there.  When you call init_edgestart, you should re-read the `INDATA` namelist.  This better compartmentalizes the code.  Also `mylocalmaster` doesn't need to be broadcast.

PR #244, opened 2024-06-04
[LIBSTELL] Use a more direct (i.e. no loops) implementation of `mpi_calc_myrange`

jonathanschilling:
This is a further iteration of #240.

> This is a fundamental and substantial change to a piece of code used by many routines.

I agree and therefore, I would say that it would be valuable to have this routine very well documented and almost obvious to understand.
In the suggested change, the assumptions and the approach are (I hope) quite explicitly documented and reasoned about. Please let me know if you think differently about this.

I also attached a stand-alone test program (`test_mpi_calc_myrange`) now that I used to verify that the computed `mystart` and `myend` are identical (for the probed parameter value ranges) to what the old implementation produced:
[test_mpi_calc_myrange.zip](https://github.com/user-attachments/files/15544282/test_mpi_calc_myrange.zip)
It is compiled with `gfortran test_mpi_calc_myrange.f -o test_mpi_calc_myrange` 
and then run with `./test_mpi_calc_myrange` and should then print `all tests passed successfully`.
I understand that it was not obvious from the previous iteration of this PR that the outputs are identical. Sorry for that.

### DIFF
--- a/LIBSTELL/Sources/Modules/mpi_params.f
+++ b/LIBSTELL/Sources/Modules/mpi_params.f
@@ -101,7 +101,7 @@ c                    differently.
       total_work = n2 - n1 + 1
 
       ! size of chunks that are present in all ranks
-      ! (Note that we _intend_ use integer division here,
+      ! (Note that we use integer division here intentionally,
       !  since the remainder is handled explicitly
       !  via the `work_remainder` variable below.)
       work_per_rank = total_work / local_size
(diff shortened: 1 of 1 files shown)


lazersos:
@jonathanschilling This looks better.  All your types are integers and so integer math is used, this could introduce issues.  I'll double check this today and get back to this.

jonathanschilling:
Thanks! The use of integer math is intended, as the remainder of the division is explicitly handled via the `work_remainder` variable. Do you know a way to make it more explicit in the code that integer division is intended? (I assume that you are concerned about the division in `work_per_rank = total_work / local_size`, right?)

EDIT: added comments in the code to preserve this discussion

lazersos (approved):

PR #245, opened 2024-06-04
[WOUT2INDATA] add a tool that puts the plasma boundary geometry from a VMEC wout file into a new INDATA namelist

jonathanschilling:
This PR adds a new tool `WOUT2INDATA`.

This tool:
* reads a VMEC `indata` namelist from a given VMEC input file,
* reads a given VMEC `wout` file and
* writes a new `indata` namelist to a new file, where the Fourier coefficients `RBC`, ... of the plasma boundary geometry have been replaced to represent the plasma boundary from the `wout` file.

The new output file is called `input.<ext>_new` and it is written with the same relative/absolute path as the given input file.


lazersos:
So I was taking a look at this and I don't quite understand this.  Wouldn't it be better to just have a utility which takes a wout file only and generates and input file from it?

jons-pf:
The problem I see is that some information is not reconstructable from the `wout` file that was used to generate it, e.g., the intermediate entries in `ns_array` but also `ncurr`, `nvacskip`, `pres_scale`, `tcon0`, `delt` and `lforbal`. We could now patch up the `wout` writing, but that still would not bring that info back for all the existing VMEC runs, for which the input files should still exist on the other hand.

jons-pf:
One could make the input file command line arg optional though and leave all parameters untouched wrt. their default values from the `vmec_input` module when generating a new input based on a `wout` file only.

lazersos:
I would also suggest the name VMEC2INDATA as this is more consistent with our existing naming conventions.

lazersos:
Also `PRES_SCALE` can be reconstructed from the `wout` file.
```fortran
REAL(rprec), EXTERNAL :: pmass
IF (pmass(0.0) > 0.0) THEN
   PRES_SCALE = presf(1)/pmass(0)
END IF
```

jons-pf:
Ok, I would then move forward as follows:
1. rename to `VMEC2INDATA`
2. make input file optional -> filename of generated new input file is then determined by `<ext>` from `wout_<ext>.nc`, since a `wout` filename always needs to be provided
3. if input file is not provided, reconstruct as much of `INDATA` as possible from `wout` (incl `pres_scale` as you suggested)
4. if input file is provided, use as much info from that as possible, i.e., overwrite only `lasym`, `mpol`, `ntor` and the boundary coefficients (as currently implemented)

@lazersos Does that address your comments to your satisfaction?

lazersos:
@jons-pf has there been progress on this?

jons-pf:
@lazersos Sorry for the delay, I was using the code as-is for now and did not get around yet to implementing your suggestions; will do ASAP.

lazersos:
I'm closing this PR becasue we now have this ability on the python side of the code, and it is more full featured. You can use `vmec_util.py --vmec EXT --wout2indata` and it will read the `wout.EXT` or `wout_ext.nc` file and output a `input.EXT_new` file. Should meet everyone's needs.

PR #249, opened 2024-06-11
Feature/beams3d eom

lazersos:
This switches the gyrocenter following version of the code over to the new form incorporating curl(B). A helper variable for the collision operators was added.  And the functional referances to nag were changed.


kudav (approved):
I think the rename makes sense and looks good to me otherwise as well.

PR #265, opened 2024-07-17
Fix typo in MPI_INFO_SET "stripping"->"striping"

cfe316:
In THRIFT and BEAMS3D there is a typo in an MPI_INFO_SET call. `striping_unit` is misspelled as `stripping_unit`. 

See for example the [MPI 3.1 standard](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report.pdf) for the correct spelling.


lazersos (approved):
Good catches.

PR #273, opened 2024-07-25
Interpolate using xrho and yrho, not S

cfe316:
Turning a position into S,U,V or visa-versa using S4D and U4D does not work well near the axis. Using X4D, Y4D where X = S cos U, Y = S sin U *also* does not work well near the axis. Instead, use XRHO4D, YRHO4D where XRHO = RHO cos U and Y = RHO sin U. XRHO and YRHO are much closer to linear functions in cartesian space, and they don't have odd behavior near the axis.
This permits finding desired (S or rho, U, phi) points which are nearly identical to the positions determined by VMEC/LIBSTELL.

* Lookups for marker scoring to the distribution function is corrected, so dense_prof is corrected
* The code for epower_prof & ipower_prof is not yet corrected
* Only init_vmec has changed; other file types' creation of S_ARR is TODO.


BEAMS3D/Sources/beams3d_init_vmec.f90
```diff
@@ -330,6 +331,7 @@ SUBROUTINE beams3d_init_vmec
          IF (ier == 0 .and. bphi /= 0) THEN ! We have field data
             ! Save Grid data
             S_ARR(i,j,k) = MAX(sflx,0.0)
+            RHO_ARR(i,j,k) = SQRT(S_ARR(i,j,k))
```
cfe316, line 334: In the three places where S_ARR is set from the VMEC file, immediately afterward I set `RHO_ARR(i,j,k) = SQRT(S_ARR(i,j,k))`.

Probably to be cleaner I should just do `RHO_ARR = SQRT(S_ARR)` once at the end.

BEAMS3D/Sources/beams3d_physics_mod.f90
```diff
@@ -305,8 +307,8 @@ SUBROUTINE beams3d_physics_gc(t, q)
             zeff_temp = max(fval(1),one)
             CALL R8HERM3FCN(ict,1,1,fval,i,j,k,xparam,yparam,zparam,&
                             hr(i),hri(i),hp(j),hpi(j),hz(k),hzi(k),&
-                            S4D(1,1,1,1),nr,nphi,nz)
-            s_temp = max(fval(1),zero)
+                            RHO4D(1,1,1,1),nr,nphi,nz)
+            rho_temp = max(fval(1),zero)
             DO l = 1, NION
```
cfe316, line 312: This should be improved further. It's generally not good enough to do a single interpolation via RHO4D. It's much better to get XRHO4D and YRHO4D, then take the sqrt to get rho.

The RHO4D function might not even have an 'axis' (loci of points where rho=0), but sqrt(XRHO4D^2 + YRHO4D^2) will since they boundary surfaces between positive and negative values.

cfe316: Same with other single S4D or RHO4D lookups in this and similar files.

cfe316 (commented):
Here's a brief description of the changes in this initial state.

lazersos:
There are some changes I'd like to make. Let's first merge this into a local branch the do a pull request into develop from there.

Issue #242, opened 2024-06-04
BEAMS3D: Radial distribution function bug


lazersos:
When BEASM3D generates distribution functions binning of particles is done in the `out_beams3d_nag.f90` and `outpart_beams3d_nag.f90` with the formula:
```fortran
d1 = MAX( MIN( CEILING( rho_help*ns_prof1 ), ns_prof1 ), 1 ) ! Rho Bin
```
This has the effect that if a particle has a value of `rho_help > 1` the code adds the particle to the ns_prof1 bin location which should be the last bin.

Two fixes to this problem are suggested:

1. Define a quantity `h1_prof` as is done for the other components and redefine so all particles which go beyond `rho=1` end up in the last gridpoint which is not used.
2. We could define an input parameter which allows the user to define where the cutoff point in rho is.

There is a final point.  Currently we use the full time of flight when a particle hits a wall but technically this should also be adjusted down so we use the correct time of flight `dt`. This requires a restructuring of `out_beams3d_nag.f90` and `outpart_beams3d_nag.f90`.

PR #274, opened 2024-07-26
Interpolate using xrho and yrho, not S

cfe316:
Getting a position using S4D and U4D does not work well near the axis. Using X4D, Y4D where X = S cos U, Y = S sin U *also* does not work well near the axis. Instead, use XRHO4D, YRHO4D where XRHO = RHO cos U and Y = RHO sin U. XRHO and YRHO are much closer to linear functions in cartesian space, and they don't have odd behavior near the axis.
This permits finding desired (S or rho, U, phi) points which are identical to the positions determined by VMEC/LIBSTELL.

* Marker scoring bin lookups have been changed, so dense_prof is corrected
* The code for epower_prof & ipower_prof is not yet corrected
* Only init_vmec has changed; other file types' creation of S_ARR is TODO.

This pull closes #274 #273 #242


lazersos:
This pull closes #274 #273 #242

cfe316:
@lazersos I'm not convinced that this addresses #242. I'm not sure if this is related or even expected, but I still see that the outermost row of (non-gutter) cells has a larger-than-expected volume. In the plot the last one or two data points are higher than the otherwise linear trend. I'd guess this has to do with the virtual casing methods, which I'm not familiar with.

Screenshot 2024-07-26 at 10 17 54 AM

Above I'm _not_ plotting the outermost (gutter) layer.

lazersos:
@cfe316 That's a fair assessment. It's not so much the virtual casing as the fact that the background grid has points which are inside the equilbirium (interpolated) and outside (extrapolated) when it comes to s (or rho). Thus the meaningfulness of the mehtod becomes more questionable at the edge of  the plasma.  There may be a way to fix this by considering values of rho slightly less than one instead of one.

lazersos:
@cfe316 OK I pushed some changes to get rid of the `S`, `X` and `Y` arrays in favor of the `RHO`, `RHOX`, and `RHOY` arrays. The code still outputs the `S_ARR` array so all interfaces are intact, the changes should all be internal. I did have to modify some parts of `beams3d_init` as we shouldn't calculate `RHO` anywhere other than in `beams3d_init` (the interfaces to EQDSK, HINT, and FIELDLINES were affected).

@kudav Can you test that W7-X and ASDEX-U slowing down runs still work as expected?

lazersos (approved):

cfe316:
@lazersos thank you for addressing this.

There's a few further points on this same topic where the code could be further improved. 
For the highest accuracy you never want to use the RHO4D interpolation at all; you probably don't even need to construct it. Like S4D it has the flaw that it's not guaranteed to have any locations with the value of S=rho=0. In contrast since XRHO and YRHO have (approximately) perpendicular planes where they are zero, there is guaranteed to be a rho=0 axis.

Just use `rho = sqrt(x^2 + y^2)` everywhere. Of course this does come with the cost of two interpolation lookups rather than one. Have you ever profiled the code to see if that's a significant cost?

In this PR so far, here are the places where a single RHO4D lookup could/should be changed to a pair of XRHO4D, YRHO4D lookups.

```
beams3d_init_fusion line 323
beams3d_physics_gc line 305
beams3d_physics_fo line 515
beams3d_follow_neut lines ~755, 815, 950
beams3d_SFLX 1922
beams3d_SUFLX 1990
```
I'm happy to make these changes if you like.

lazersos:
@cfe316 I have mixed feelings about this. If S and U are both needed then we be replaing two spline evaluations with two spline evaluations so that's a wash. S = XRHO*XRHO+YRHO*YRHO so if we really only need S then we're only adding two multiplies and an addition. U most likely requries a call to ATAN2(YRHO,XRHO) and a MODULO. If RHO is actually required then we need a SQRT(S) as well.

Fundamentally, I'd  really like to get rid of RHO, U and S in the code after the `beams3d_init` routine. But in order to evaluate 'profile' quantities we need it. There is the additional distribution function issue as well. We should really be storing the distribution function on an R,phi,Z grid but such a grid is super wasteful in terms of storage for a stellarator as most of the grid is outside the domain of interest. Additionally, the distribution function is 6D in our formulation. Finally, the distribution function for NBI and ICRH violates field periodicity requring a full torus grid in phi.

Long story short, let's save turn this idea into separte issue and wait for @kudav to get back to us on whether W7-X and ASDEX-U runs still produce something meaningful.

kudav:
grafik
The AUG run on Raven doesnt want to start, the W7X case does seem to be meaningful but there is a discrepancy in the core which might just be caused by the different/more accurate volume normalization. However, the dip on the axis still persists it seems.

cfe316:
>  S = XRHOXRHO+YRHOYRHO so if we really only need S then we're only adding two multiplies and an addition. U most likely requries a call to ATAN2(YRHO,XRHO) and a MODULO.

Isn't the `_4D` spline lookup at least evaluating a few cubic equations? Is a single SQRT or ATAN2 actually much slower because of the more complicated algorithms?

Also: S = XRHO^2 + YRHO^2 at the spline knots, but nowhere else. That's why it's important to use XRHO4D and YRHO4D near the axis.

As one slightly crazy idea: if you just need S or rho
1. Do a single call to RHO4D
2. If rho < 0.1 then ditch that result; use the full XRHO4D and YRHO4D. This would be more expensive but only in 1% of the calls.

@kudav Are you plotting the dense_prof output? Or your own sum over dist_prof?

lazersos:
@cfe316 You are correct. But anytime we've reference the underlying grid we only do it at the knots if at all I believe.  However, if `X` and `Y` are the splined quantities then `S=X*X+Y*Y`.

Regarding the splines, we usually handle all the coefficient calcs outside the spline evaluation so in the end the spline evaluations are just additions and multiplications. Hence the variables `xparam` etc. that we calculate right before we evaluate a spline. `SQRT` isn't terrible but more costly than two multiplies and an addition. The `ATAN2` is pretty expensive though from what I can find.

PR #281, opened 2024-08-22
B3D: output of calc_birth + att. descriptions

kudav:
This is a minor bugfix adding the calc_birth flag to the fidasim namelist by beams3d and also correcting some descriptions and units.


lazersos (approved):
Approved.

PR #283, opened 2024-09-18
VMEC <B.B> axis bugfix

### LINKED ISSUE #282: VMEC2000: Extrapolation of bdotb and jdotb
In `VMEC2000/Sources/Input_Output/jxbforce.f`, the linear extrapolation of `bdotb` towards the magnetic axis is not correct. Currently it is coded as

`bdotb(1) = two*bdotb(3) - bdotb(2)`

while it should be

`bdotb(1) = two*bdotb(2) - bdotb(3)`

as is done for the other variables: `jdotb` and `bdotgradv`. 

With the bug, the profile of <B.B> looks like:

Screenshot from 2024-09-18 11-28-39

Once the bux is fixed the profile will look like:

Screenshot from 2024-09-18 11-36-07

Regarding `jdotb`, we must re-think if this way of extrapolating is the most appropriate one, since doing this linear extrapolation may result in a <J.B> profile as such:

Screenshot from 2024-09-18 11-47-58

lazersos:
This pull request closes #282 by applying the suggested fix.  It also adds plotting of `bdotb` and the Mercier quantities to VMECplot.


ajchcoelho (changes requested):
Why in  `self.ax.plot(self.nflux,self.vmec_data.bdotb/1000)` B-squared is divided by 1000?

ajchcoelho (approved):

PR #284, opened 2024-10-04
Merge PENTA3 from feature/ORNL_PENTA to develop branch

ajchcoelho:
I suggest that not only PENTA is merged, but also:

- ``DKES`` -- I created a new version of dkes_input_prepare.f in order to be able to run several cmul, efield when reading from the input namelist; I've also changed the ``krun`` parameter to allow a maximum of 300 runs (previously it was set to 100)

- ``pySTEL`` --  I've changed the collisions class (fixed some issues and added new functions) and created 3 new classes:  dkes, plasma and penta. These are useful to read output from DKES and PENTA, as well as to create all the necessary input files to run these codes


lazersos (approved):
Looks good to me.

PR #289, opened 2024-10-11
Feature/py stel

lazersos:
This pull request merges in python utility improvements into the main develop branch. Only minor modifications are made to a couple modules in LIBSTELL.


pySTEL/libstell/fusion.py
```diff
@@ -43,6 +42,59 @@ def __init__(self):
...
+		B0 : real
+			Average magnetic field [T]
+		iota : real
+			Rotational transform at 2/3 flux [arb]
+		Returns
+		----------
+		tauiss04 : real
+			Confinement time [s]
+		"""
+		return 0.134*a**2.28*R**0.64*P**-0.64*navg**0.54*B0**0.84*iota**0.41
```
ajchcoelho: Exponent of Power is -0.61.

pySTEL/libstell/fusion.py
```diff
@@ -43,6 +42,59 @@ def __init__(self):
...
+		"""Sudo Density Limit
+
+		This routine computes the Sudo density limit scaling as defined
+		in:
+		https://doi.org/10.1088/0029-5515/30/1/002
+
+		Parameters
+		----------
+		P : real
+			Total Power [MW]
```
ajchcoelho: Absorbing Power is in W since it's already divided by 1E6 below.

ajchcoelho (changes requested):
There are two typos in fusion.py

pySTEL/libstell/fusion.py
lazersos: Corrected these two issues in the recent push.

kudav:
I finally got around to set up pySTEL and I think I stumbled across a bug: python (h5py) and matlab/fortran seem to have different standard major directions when reading and writing hdf5 files, see e.g. [here](https://stackoverflow.com/questions/24116713/hdf5-rowmajor-or-colmajor) and [here](https://github.com/JuliaIO/HDF5.jl/issues/785). This means that right now the calculations run through without error for nr=nz, i.e. in beams3d.calcVolume. When nr!=nz, the calculations fail. 
I'm using python version 3.12.4 and h5py version 3.11.0, and hdf5-mpi/1.12.2 for libstell, although the output files i'm looking at might have been created with different versions (shouldnt matter since I'm normally analyzing them with always the same matlab scripts).

lazersos:
@kudav Can you point me to an HDF5 which I can use for testing.

lazersos:
@kudav I've made some changes and it should work now, but it's worth spending some time to check against the Matlab routines to make sure the numbers are consistent. The BEAMS3D support is still a work in progress.

PR #290, opened 2024-10-18
Feature/thrift

ajchcoelho:
Modifications in `THRIFT`:

- Corrected the labels of `phip_spl` and `vp_spl`; the 1st one is actually derivative with respect to rho while the second one is derivative with respect to toroidal flux (check `load_vmec.f90`)
- Quantity <E.B> is evaluated and saved in `THRIFT_EPARB`

Modifications in `pySTEL`:

- Added function to the classes `PLASMA`, `PENTA` and `THRIFT`


lazersos (approved):
Lookd good to me.

PR #293, opened 2024-10-30
Merge of feature/thrift_restart into develop and feature/THRIFT_DKES_PENTA

ajchcoelho:
A restart mode to THRIFT is introduced, meaning that now we can do a long simulation in several runs. The code reads a restart file whose name is specified in `&THRIFT_INPUT` under `RESTART_FILENAME`. The restart file can be the results file of the previous run. The boolean `RESTART_FROM_FILE` should be set to `TRUE` in the input namelist  in case the restart file is to be read, otherwise the code will ignore the restart mode and start a simulation as usual.

I have tested this new functionality in the following way. I took the input files in `BENCHMARKS/THRIFT_TEST` and made a single run from `tstart=0.0` to `tend=80.0`. Then I made multiple runs using restart files in the intervals `t=[0,20]`, `t=[20.5,40]`, `t=[40.5,60]` and `t=[60.5,80]`. The variable `THRIFT_UGRID` is compared at different times for the two cases (single run vs multiple runs):

image
image
image

Note that for any `t` in the interval `[0,20]` , the difference between single run and multiple runs should be 0.0 since both cases are essentially starting from the beginning and doing the exact same thing. After `t=20`, the divergences are of order `1E-10` which is probably due to the way the restart file is being read.

Finally, I have also tested this using a SQUID wout file and the bootstrap computed with `BOOTSJ` . The comparison _single run vs multiple runs_ gives similar results, i.e., differences of order `1E-10`.

### DIFF
--- a/THRIFT/Sources/thrift_init.f90
+++ b/THRIFT/Sources/thrift_init.f90
@@ -212,11 +212,6 @@
             DEALLOCATE(temp2d,temp1d)
 
          END IF
-
-         ! Bcast tstart
-         CALL MPI_BARRIER(MPI_COMM_SHARMEM,ierr_mpi)
-         CALL MPI_BCAST(tstart,1,MPI_DOUBLE_PRECISION,master,MPI_COMM_SHARMEM,ierr_mpi)
-         IF (ierr_mpi /= MPI_SUCCESS) CALL handle_err(MPI_ERR,'set tstart',ierr_mpi)
       END IF
 
       ! Check that tend > tstart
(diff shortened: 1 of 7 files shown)


lazersos:
In keeping with code convention, I'd suggest that instead of specifying the restart file in the `THRIFT_INPUT` namelist we do it as a command line option. This keeps in convention with other codes. I'll make the mods to the branch and push the changes.

lazersos (approved):
Looks good.

PR #300, opened 2024-11-04
Spell "beams3d" correctly on its docs page

cfe316:
Fix common typo.


lazersos:
@cfe316 In the future you don't have to make a pull request to update the gh-pages. Generally, these are community documents and it is assumed that changes here won't 'break' anyone elses work.  But good catch none the less.

cfe316:
Thanks, will do!

PR #301, opened 2024-11-06
Feature/stellopt qi

lazersos:
This pull request adds the QI target as defined in (http://dx.doi.org/10.1103/PRXEnergy.3.023010) by Alan Goodman _et al._ There is one subtly that we don't inlclude the `B_I dot grad(phi)` term found in the denominator of the `dl` term.


lazersos:
As there hasn't been any review of this in two weeks and it generally seems like it won't break anything (along with having a few bugfixes). I'm going to merge without review.

PR #304, opened 2024-11-14
Merge THRIFT_DKES_PENTA into develop

ajchcoelho:
The integration `DKES+PENTA+THRIFT` is fixed and ready to merge. There are still some features I'd like to add, but they are not very important so I will leave them for the future. 

To make sure the integration is correct, I have ran THRIFT with `bootstrap_type='dkespenta'`. I  ran THRIFT 1 time step which allows to obtain J_BS at t=0, which can then be compared with the result of running PENTA by itself:

image

Note that J_BS from `DKES+PENTA+THRIFT` is the result of an interpolation at THRIFT's grid, hence why the maxima and the minima. Nevertheless, the two curves agree at the input surfaces (which were set as the same when running THRIFT and PENTA).


lazersos (approved):
Changes look good to me.

PR #307, opened 2024-11-16
BEAMS3D: Ability to mark self-generated markers with population index

lazersos:
This added the capability to use `DEX_BEAMS` to put each user defined marker in a separate population. Useful for quickly generating heating benchmark plots. If DEX_BEAMS is not specified the code should behave like it always did placing all user generated markers in one population.

### DIFF
--- a/BEAMS3D/Sources/beams3d_init.f90
+++ b/BEAMS3D/Sources/beams3d_init.f90
@@ -762,10 +762,21 @@
          charge = charge_in(1:nparticles)
          mu_start = mu_start_in(1:nparticles)
          t_end = t_end_in(1:nparticles)
-         beam  = 1
-         nbeams = 1
-         charge_beams(1) = charge_in(1)
-         mass_beams(1)   = mass_in(1)
+         IF (nparticles <= MAXBEAMS) THEN
+            beam(1:nparticles)  = Dex_beams(1:nparticles)
+            charge_beams(1:nparticles) = charge(1:nparticles)
+            mass_beams(1:nparticles) = mass(1:nparticles)
+         ELSE
+            beam(1:MAXBEAMS)  = Dex_beams(1:MAXBEAMS)
+            beam(MAXBEAMS+1:nparticles)  = Dex_beams(MAXBEAMS)
+            charge_beams(1:MAXBEAMS) = charge(1:MAXBEAMS)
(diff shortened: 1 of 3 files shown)


pySTEL/pyproject.toml
kudav, line 1: This does not seem to belong here, but would rather fit in a pySTEL branch?

BEAMS3D/Sources/beams3d_init.f90
kudav, line 1: This logic only allows setting the dex_beams of up to 32 particles, which could cause confusion. I'd suggest increasing MAXBEAMS or at least having dex_beams be longer than MAXBEAMS. I think it might be a good idea to have something separate like dex_beams_prt to avoid confusion with the usual dex_beam behavior.

pySTEL/pyproject.toml
lazersos: Yes, there's a separate pull request.

BEAMS3D/Sources/beams3d_init.f90
lazersos: The issue is that having a large number for MAXBEAMS would present the possibility of creating huge distribution arrays. This is really a feature for grouping particles into distributions not so much to provide a distribution function for each particle.

kudav:
I think the way it is right now results in some unexpected behavior. I would like to run a slowing-down benchmark with 50 particles with individual pitch values and group these together to have eg individual heating profiles. Loading them in works fine, but using dex_beams is restricted to the first 32 particles:
Running with
DEX_BEAMS = 31*1  2 
produces a run with 50 particles, with the first 31 in the first bin and the following particles in the second bin, for which there is no individual control

Running with
  DEX_BEAMS = 31*1  2  2
Produces the following error:

----- Input Parameters -----
ERROR reading namelist BEAMS3D_INPUT from file: 
input.ORBITS_slow
Invalid line in namelist:   DEX_BEAMS = 31*1  2 2

Which does not help to resolve the underlying issue.

My point is, dex_beams for particles fulfills a different purpose (grouping potentially many markers in a few bins) than for the beamlet model (assigning BEAMS3D beams a beamlet index).

lazersos:
Just change `MAXBEAM=50` for your case and recompile. Restructuring the code would be required at this point and I don't want to do that. This is a simple fix to allow individuals to run an edge case.

PR #308, opened 2024-11-18
Bugfix/stellopt dkes alpha

lazersos:
Fixes a bug where the headers for DKES_ALPHA didn't print the correct number of lines.


leebr48 (approved):
Looks good.

PR #312, opened 2024-11-27
Feature/thrift vmec restart

lazersos:
This pulls request does the following:

- Merges changes from [ORNL-fusion/PARVMEC](https://github.com/ORNL-Fusion/PARVMEC)
- Fixes VMEC restart feature to work when more than one processor is requested
- VMEC restarting is implemented in THRIFT.


THRIFT/Sources/thrift_paraexe.f90
ajchcoelho, line 1: I think it's better to write line 131 as:

`IF (lhit .and. .not. mytimestep==1) reset_string ='wout_reset_file.nc'`

ajchcoelho (changes requested):
When running this branch with the input files in BENCHMARKS/THRIFT_TEST (`profiles_analytical.h5` and `input.analytical`), the code crashes with a segmentation fault error at `t=12.5s`, `substep=12`.

I tracked it down to line 38 of file `thrift_run_equil.f90`. The `phi` variable suddenly is not allocated. I don't understand how it runs for more than 12 time iterations and then this error pops up.

THRIFT/Sources/thrift_paraexe.f90
lazersos: Implemented this.

lazersos:
I've added a feature to allow the user to control use of the VMEC 'reset' feature `-vmec_reset`.

lazersos:
The issue you're experiencing is because the VMEC equilibria is failing to converge. At which point `ier_paraexe=-1`, which we don't handle. From that point on the code basically stops calculating VMEC equilibria which eventually leads to a crash. A check needs to be implemneted so that once VMEC fails to converge, the code exits out with the proper message.

lazersos:
The code has been updated so that if VMEC fails to converge in `reset` mode then it'll attempt a second run in the normal mode of operation.

lazersos:
@ajchcoelho I've updated the code so that the `wout_reset_file.nc` files are only written after THRIFT successfully converges on a total current. This seems to fix the issue I had where at `t=2.5 s` in the BENCHMARKS case the code would seemling fail to converge the equilibria. The idea being here that `wout_reset_file.nc` should always be the previous timestep not the previous iteration. I also added some initializations which VMEC wasn't doing.

ajchcoelho (approved):
Looks good. I've made a comparison for the analytical case, with restart and w/o restart at different times:

image

In terms of timing, as already expected there is no time improvement: THRIFT spends 25 minutes running the equilibrium code (VMEC) in both situations (with and without restart mode)

PR #316, opened 2024-12-02
Feature/beams3 d coll update

kudav:
- Minor bugfixes for the nubeam collision operator, which for now remains commented out.
- Added velocity diffusion to slowing down and heating calculation in accordance with [this paper](https://www.sciencedirect.com/science/article/pii/002199918190111X).


lazersos:
@kudav The new velocity diffusion operator violates energy conservation in the code. Markers can now gain and lose energy randomly. Before this feature can be added to the repo I'd need to see a comparrision for a W7-X slowing down case that the ion and electron heating are not significnatly impacted by this.

kudav:
Here is the heating comparison for a W7X case with the current develop branch and the requested changes:
grafik

lazersos:
@kudav OK, was that with 64k markers per energy?  That looks supprisingly noisy.  I'll run the benchmark case and post the results.

kudav:
Yes, 64k markers per energy with two active beams. I think the structure around half radius comes from the iota=1 crossing there.

lazersos:
From branch `develop`
alpha_heating_example

From branch `feature/BEAMS3D_coll_update`
alpha_heating_diffusion

These look good. I want to check that this hold for more W7-X like parameters.

lazersos:
OK and for more experimental device like conditions (4 keV electron temperature).

`develop`
proton_heating

`feature/BEAMS3D_coll_update`
proton_heating_diffusion

Clearly there are more issues with lower energy particles and systems. Clearly this is a negligible effect for reactor systems but can have a significant influence on experimental systems. Perhaps we include the new effect but wrapper it with pre-processor directives.

lazersos:
The wrappering of the pre-processor directives has been handled in the other pull request #317.

lazersos:
It should now be up to date with the `develop` branch.

lazersos (approved):
We should probably add a section on the compilation page of the wiki about pre-processor flags.

PR #317, opened 2024-12-02
Feature/beams3 d toroidal rotation

kudav:
This adds the toroidal rotation to the collisional calculation to BEAMS3D, both for deposition and slowing down. 
New input namelist parameters are OMEG_AUX_S and OMEG_AUX_F, giving the radial profile of the angular rotation velocity. 
The toroidal velocity changes the relative velocity (i.e. the parallel velocity) in the collision calculation, which takes place in the particle frame.
Note that the sign of the rotation is positive in the direction of the magnetic field, not in the geometric toroidal angle!

The orbit effects of toroidal rotation can be incorporated by including the equivalent radial electric field.


BEAMS3D/Sources/beams3d_init_continuegrid.f90
```diff
@@ -86,7 +87,7 @@ SUBROUTINE beams3d_init_continuegrid
    CALL MPI_CALC_MYRANGE(MPI_COMM_LOCAL, 1, nr*nphi*nz, mystart, myend)
 
    IF (mylocalid == mylocalmaster) THEN
-      TE = 0; NE = 0; TI=0; S_ARR=1.5; U_ARR=0; POT_ARR=0; ZEFF_ARR = 1;
+      TE = 0; NE = 0; TI=0; S_ARR=1.5; U_ARR=0; POT_ARR=0; ZEFF_ARR = 1;OMEG_ARR = 1;
```
lazersos: Shouldn't `OMEG_ARR=0` here?

BEAMS3D/Sources/beams3d_init_eqdsk.f90
```diff
@@ -152,7 +153,7 @@ SUBROUTINE beams3d_init_eqdsk
       CALL MPI_CALC_MYRANGE(MPI_COMM_LOCAL, 1, nr*nz, mystart, myend)
 
       IF (mylocalid == mylocalmaster) THEN
-         TE = 0; NE = 0; TI=0; S_ARR=1.5; U_ARR=0; POT_ARR=0; ZEFF_ARR = 1;
+         TE = 0; NE = 0; TI=0; S_ARR=1.5; U_ARR=0; POT_ARR=0; ZEFF_ARR = 1;OMEG_ARR = 1;
```
lazersos: Shouldn't `OMEG_ARR=0` here.

BEAMS3D/Sources/beams3d_init_fieldlines.f90
```diff
@@ -87,7 +88,7 @@ SUBROUTINE beams3d_init_fieldlines
       CALL MPI_CALC_MYRANGE(MPI_COMM_LOCAL, 1, nr*nphi*nz, mystart, myend)
 
       IF (mylocalid == mylocalmaster) THEN
-         TE = 0; NE = 0; TI=0; S_ARR=1.5; U_ARR=0; POT_ARR=0; ZEFF_ARR = 1;
+         TE = 0; NE = 0; TI=0; S_ARR=1.5; U_ARR=0; POT_ARR=0; ZEFF_ARR = 1;OMEG_ARR = 1;
```
lazersos: Shouldn't `OMEG_ARR=0` here?

BEAMS3D/Sources/beams3d_physics_mod.f90
lazersos, line 1: This needs to be redone. In the end all you want to do is:
`vll_temp = vll_temp - omeg_temp*r_temp`
The way this was implmented needs to be drastically chagned.
- Remove the call to `beams3d_MODB` this is handled below in the statement block.
- Remove `beams3d_lab_to_plasma` and `beams3d_plasma_to_lab`
- Add the spline evaluation to the big block of spline evaluations.
- Add the modification to vll to the top of the helpers block `vll = vll + omeg_temp*r_temp`
- At the end of the `beams3d_physics_gc` subroutine add the velocity modification back in. `q(4) = q(4) - omega_temp*r_temp`.

BEAMS3D/Sources/beams3d_write.f90
```diff
@@ -145,6 +145,10 @@ SUBROUTINE beams3d_write(write_type)
                   CALL write_var_hdf5(fid,'ZEFF_ARR',nr,nphi,nz,ier,DBLVAR=ZEFF_ARR,ATT='Effective Ion Charge',ATT_NAME='description')
                   IF (ier /= 0) CALL handle_err(HDF5_WRITE_ERR,'ZEFF_ARR',ier)
                END IF
+               IF (ASSOCIATED(OMEG_ARR)) THEN
+                  CALL write_var_hdf5(fid,'OMEG_ARR',nr,nphi,nz,ier,DBLVAR=OMEG_ARR,ATT='Toroidal rotation [m/s]',ATT_NAME='description')
```
lazersos: `ATT` is wrong it is in units of `rad/s`.

BEAMS3D/Sources/beams3d_write_ascoth4.f90
lazersos, line 1: Doesn't look like you actually output omega_I based on your modifications.

LIBSTELL/Sources/Modules/beams3d_input_mod.f90
```diff
@@ -153,6 +154,8 @@ SUBROUTINE init_beams3d_input
       ZEFF_AUX_F = -1
       POT_AUX_S = -1
       POT_AUX_F = -1
+      OMEG_AUX_S = -1
+      OMEG_AUX_F = -1
```
lazersos: `OMEG_AUX_F = 0`

LIBSTELL/Sources/Modules/beams3d_input_mod.f90
lazersos, line 1: You should also make sure this new variable is written out in the namelist write routine.

lazersos (changes requested):
Major changes necessary.

BEAMS3D/Sources/beams3d_physics_mod.f90
lazersos, line 1: OK, I'm being a bit fussy here but the use of `omeg_temp` is inconsistent betweeen `beams3d_physics_gc` and `beams3d_physics_fo`. In `beams3d_physics_gc` the variable is treated as the parallel velocity while in `beams3d_physics_fo` it modifies `vll` but looks like `v_toroidal`. 

I think it's safe to assume that `omeg_temp` should be treated as the toroidal rotation. So the 329 in `beams3d_physics_gc` should read
```fortran
vll = vll - omeg_temp*r_temp*bphi_temp/modb
```
And later on as well.  This requires the quantity `BPHI4D` be splined as well.
In `beams3d_physics_fo` the modification should not be made to `vll` but rather `q(5)` as this is toroidal velocity.

lazersos (changes requested):
There is still a slight issue with how the velocity is handled.

lazersos:
I've made a bunch of changes so that the rotational velocity is handled better. Additionally, the multi-pass deposition option breaks the benchmark becasue there is no explicit wall.

lazersos:
Multi-pass is now enabled but only for non-plasma_only runs.

lazersos (approved):
From my point of this branch is now ready to merge.

kudav:
Note that due to the changes OMEG is now in the toroidal direction (no longer parallel to the toroidal field) and the sign is made parallel to the magnetic field by multiplication with BPHI. This introduced a small bug in the fidasim interface inconsistent with this new behavior, which I've adressed in #319.

PR #319, opened 2024-12-05
Feature/b3 d continue pot

kudav:
This extends the continuegrid functionality to be able to read the existing background profile arrays (if available) and use those for collisions.
If profiles are specified in the input namelist, they take precendent and overwrite the existing ones. This allows the direct import of the background plasma from codes able to output all the quantities on a R-Z grid (like Jorek). It also allows flux-label independent profiles to be used if desired, e.g. for SOL simulations.

Note that currently the ion density is not read in from the original file, which should be adressed before merging.


lazersos:
Shouldn't the ability to 'continue' a run only read values from the HDF5 file and not an input namelist?

kudav:
I think having the option to use profiles controlled by the input namelist adds flexibility and is easy enough to control by having the profile information there or commenting it out. At least in cases where only S and U are included in the continue file, the profile information in the input namelist is necessary. We could have the continue file take precedent over the input profile namelist, but I think the option should be there to change the profiles to whatever is in the namelist.

lazersos:
I'm going to close this and not merge. The general idea behind this feature is to allow grids quantities to be read from a HDF5 file instead of recomputing them. This is fine and make sense to implement.  It should be implemented to override all INDATA namelist options, otherwise how the code works becomes WAY too confusing. Additionally, it should be independent of the particle initialization. I think this branch has promise but the way it's implemented is too messy to merge.  The changes are:

- Should only dictate where the background fields come from (no particle stuff).
- Should fully override whatever is in the input namelist.

PR #320, opened 2024-12-09
Feature/thrift_improve

ajchcoelho:
2 major modifications to `THRIFT`:

--> In case `PENTA` fails to find an Er-root, the range of search is automatically expanded  (thus far the program stopped if it wasn't able to do find any root)

--> All the ambipolar roots, including Er and neoclassical fluxes, are saved in a file (this is optional and controlled by a flag in the input namelist)

Then there are minor modifications to the classes `plasma` and `thrift`, and a new class is introduced -- `popcon` -- which allows to construct POPCON plots using the `plasma` class abundantly.

### DIFF
--- a/PENTA/Sources/penta_interface_mod.f90
+++ b/PENTA/Sources/penta_interface_mod.f90
@@ -1158,7 +1158,7 @@ MODULE PENTA_INTERFACE_MOD
       RETURN
    END SUBROUTINE penta_run_4_ambipolar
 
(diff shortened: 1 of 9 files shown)


lazersos (approved):
Looks good. We may want to keep the VMEC `wout` files for future analysis but probably OK to delete them for now.  Also we should probably create a `popcon_util.py` routine for helping users to just use the class quickly.

PR #326, opened 2024-12-18
Feature/collision refactor

### LINKED ISSUE #323: Get Collision Operators into LIBSTELL
The collision operators in `BEAMS3D` should be part of `LIBSTELL` such that they can be easily used by other codes (such as `THRIFT`).

lazersos:
This pull requests moves the computation of various coulomb logarithms and Spitzer collisional time to a LIBSTELL based module so that both BEAMS3D and THRIFT can refernece a common database.  The BEAMS3D regression tests all pass so there should be no chage BEASM3D performance.

@ajchcoelho will handle interfacing the new module into THRIFT.

### DIFF
--- a/BEAMS3D/BEAMS3D.dep
+++ b/BEAMS3D/BEAMS3D.dep
@@ -84,6 +84,7 @@ beams3d_follow.o: \
       ../../LIBSTELL/$(LOCTYPE)/stel_kinds.o \
       ../../LIBSTELL/$(LOCTYPE)/safe_open_mod.o \
       ../../LIBSTELL/$(LOCTYPE)/wall_mod.o \
+      ../../LIBSTELL/$(LOCTYPE)/collision_operators.o \
       ../../LIBSTELL/$(LOCTYPE)/mpi_inc.o \
       beams3d_runtime.o \
       beams3d_lines.o \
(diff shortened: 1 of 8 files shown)


kudav:
Shouldn't we include the other coulomb logarithms and collision parameters (coll_op_nubeam) as well?

lazersos:
> Shouldn't we include the other coulomb logarithms and collision parameters (coll_op_nubeam) as well?

When a peer reviewed publication on NUBEAM exists where they properly describe the derivation of their operator and verify the behavior, then I think it's worth considering. For now it's in BEAMS3D to support the comparrision on ASDEX. The issue is that no device/diagnostic has enough fidelity to proivde a validation case to allow comparrision between different collisional models for fast ions. Additionally, this module is to support broader transport work and just happens to overlap a bit with BEAMS3D, hence the changes.

kudav:
#319 has some bugfixes in commits [9163fc4](https://github.com/PrincetonUniversity/STELLOPT/pull/319/commits/9163fc4186866711fe70bbcf5c1cc57ec08efe9f) [a7b8c7d](https://github.com/PrincetonUniversity/STELLOPT/pull/319/commits/a7b8c7dedd45aed38ee5084e212a6a6659bdbfbf) and [1440758](https://github.com/PrincetonUniversity/STELLOPT/pull/319/commits/144075860dbc73d6e400e85e74caaa93be744cbe). Can we put them in here so they are fixed in the new version?

lazersos:
@kudav If you have bugfixes, they should probably go in separate pull requests from the feature pull request.  Create a branch based on develop and cherrypick each bugfix as a separate branch and make separate pull requests.

kudav (approved):
There was a bug in fact_crit which resulted in a critical velocity about two orders of magnitude too large because sqrt(electron_mass)^-1 was used instead of sqrt(plasma_mass)^-1. I discovered this when my slowing down run with the NRL calculation took only a minute (instead of an hour), by which time all particles had slowed down.
I also confirmed that fact_crit and fact_crit_weiland were the same save for the Z_plasma factor and removed the additional calculations. This also makes the difference between the two clear.

lazersos (commented):
Looks good.

PR #327, opened 2024-12-20
LIBSTELL: Bugfix for the particle swarm optimization.

lazersos:
This is a quick patch for a bug which prevented the particle swarm optimization from correctly identifying minimums and would result in a crash of the code at the end of optimization.


ajchcoelho:
Looks good.

ajchcoelho (approved):

PR #335, opened 2025-01-07
Bugfix/b3 d thermalization

kudav:
Stability fixes to the thermalization calculation after the recent changes incorporating velocity diffusion.


BEAMS3D/Sources/beams3d_physics_mod.f90
```diff
@@ -285,7 +286,7 @@ SUBROUTINE beams3d_physics_gc(t, q)
             CALL R8HERM3FCN(ict,1,1,fval,i,j,k,xparam,yparam,zparam,&
                             hr(i),hri(i),hp(j),hpi(j),hz(k),hzi(k),&
                             MODB4D(1,1,1,1),nr,nphi,nz)
-            modb = fval(1)
+            modb = max(fval(1),zero)
```
lazersos: MODB4D is by deffinition positive deffinite. If not then there is something wrong with the magnetic field. I'd prefer to leave this line like it is.

lazersos (changes requested):
One minor change then we can go ahead and merge.

lazersos (approved):

PR #339, opened 2025-01-27
Feature/thrift print dkes coeffs

ajchcoelho:
- Added new feature in `THRIFT`: print DKES coefficients in file at each time iteration. Will only do so if `save_DKES_coeffs` is set to `TRUE` in `THRIFT_namelist`.
This will allow us to assess _a posteriori_  how much the coefficients change between iterations.

- This pull request also includes minor changes in `PENTA` and `pySTEL` (mostly cleaning and small improvements).


lazersos (approved):
Looks good from my end.

PR #345, opened 2025-02-05
BEAMS3D: Removed beam multi-pass option

### LINKED ISSUE #332: LSODE issues: particles outside wall?
Having two-or-three issues with Beams3D. I'm not sure how they might all be connected, or not.

* I'm running on `perlmutter` using version "4.07".

I'm trying to do deposition and slowing runs on a W7X file. My immediate issue is that during the slowing runs I'm getting LSODE issues and finally crashing. 

The LSODE issues are something to do with NaNs, shown here:
```
----- FOLLOWING GYROCENTER TRAJECTORIES -----
       Method: LSODE
   Particles:      2752
       Steps:    500001   dt_min:   23.461E-09   dt_max:  100.000E-09
      NPOINC:         3
         Tol:   10.000E-09  Type: 10
 ------------------
               269        1538        1538          10
               269           4   3.6338588847276410E-010   8.1317076423414774E-008
               269   5.2021317277503769        1.5628771011341376       0.49176378049793323                            NaN                       NaN
               269           2   1.0000000000000000E-008   1.0000000000000000E-008   1.0000000000000000E-008   1.0000000000000000E-008   1.0000000000000000E-008
               269           1          -3           0
               269   0.0000000000000000        0.0000000000000000        0.0000000000000000
               269           0           0           0
               269   1.6147920165610976E-011   6.1599426901286987E-017   3.6338588847276410E-010                       NaN
               269           0           0           0           2           2           4          84          20
 ------------------
 !!!!! ERROR !!!!!
   BEAMS3D ENCOUNTERED AN LSODE ERROR
   ```
   
   There is also 
   
   ```
     BEAMS3D ENCOUNTERED AN LSODE ERROR
      CALLING FUNCTION beams3d_follow_gc
      IERR:                -3   
  ```
  and finally traces like
  ```
#2  0x7fd335653dbf in ???
#3  0x57c5f9 in out_beams3d_gc_
        at ../Sources/out_beams3d_gc.f90:69
#4  0x5b3a6b in beams3d_follow_gc_
        at /global/homes/j/jschwart/sw/STELLOPT/BEAMS3D/Debug/beams3d_follow_gc.f90: 286
...
```
   
I'm now looking at the .h5 file which is the output of the initial deposition step and seeing some unexpected outputs.

Unexpected result 1
---------------------
Some markers are at the boundary upper and lower boundary box edges: how did they get there?
See the blue points in the image.

image

In the image, the beam lines are shown at right. I don't have a working wall file so I authored this 'rectangular torus' wall file and cut out three of the triangles so the beams can make it into the plasma.
I'm not sure how some of the markers at center and left got there --- did they need to pass through the walls? Did they start with a large random directional scatter?
(I have `NPOINC=3` and I am plotting the second locational index.)

Unexpected results 2
---------------------
I would have expected that markers had positions (gyropositions?) inside the LCFS, indicated by s < 1. However, 7 particles have a neutral ionization point with 1 < S < 1.02, and many more have gyrocenter S in the same range or even up to 1.5:

```
> s_lines[np.logical_and(s_lines[:,2] > 1, end_state[:] == 0), 2] 

# particles with gyrocenter position S > 1 and which have their end state == 0 ("orbiting"); I've also excluded the NaN-cases mentioned below.

array([1.06066286, 1.00356287, 1.00689848, 1.02347276, 1.00956314,
       1.03393613, 1.02394802, 1.01666715, 1.03470434, 1.11131292,
       1.02210884, 1.07166826, 1.10615378, 1.09248667, 1.00797324,
       1.02235147, 1.00348526])
```
image
First pic: markers with s > 0.

image
Second pic: there's light between the points and the LCFS.

Unexpected result 3
--------------------
I was surprised to find that I couldn't load some Datasets of the HDF5 file in Mathematica, related to particle positions, because they had `nan`. I haven't done a full investigation, but a set of 62 particles has, for its _third_ positional index (the initial position of the **gyrocenter**) `nan` for at least `R_lines`, `U_lines` and `vll_lines`. These particles are located within the plasma itself, within the same region as the rest of the normally ionized particles.
See the images here, where a few of them are shown in red:
image
image

Questions on a way forward
----------------------------
Here's a few things I'm considering:

* Programmatically edit the `.h5` file from step 1 (deposition) to remove particles with `nan`-positions or which have `s > 1`. The latter may be important since, at the moment, I'm using the LCFS as the wall in Step 2 (slowing).
  * Also remove the stray particles at the simulation boundary.
* Author a wall file which is like the VMEC boundary but slightly extended in R(\theta, \phi) - R_axis(\phi) and Z(\theta, \phi) in order to encompass the stray `s > 1` particles from Step 1.

Might one of these work? Has either been done before? Is there a better way to cull bad particles before Step2 or to make step2 more resilient?

lazersos:
The pull request removes the multi-pass beam option from BEAMS3D. The multipass option was implemented by @kudav to deal with ASDEX-U tangential beams which 'scrapped' the inner stack. Here the beamline would allow some particles which are not ionized on the first pass to reenter the plasma from the inboard side. Unfortunately the implementation failed to properly catch the wall in some cases for shine-through particles. In particular @cfe316 demonstrated that particles for more radial beams would often find themselves on the other side of the torus before ionizing or being detected as wall hits. For this reason the modifications are being removed.

The ability to catch re-entrant particles is a worthwhile modificaiton to the code but must be done more rigerously. An issue related to this topic with a proposed test problem (not ASDEX-U) will be generated to properly develop and test such a feature. 

This should close #332

### DIFF
--- a/BEAMS3D/Sources/beams3d_physics_mod.f90
+++ b/BEAMS3D/Sources/beams3d_physics_mod.f90
@@ -743,7 +743,6 @@ MODULE beams3d_physics_mod
          !     Local Parameters
          !--------------------------------------------------------------
          INTEGER, PARAMETER :: num_depo = 256
-         INTEGER, PARAMETER :: max_beam_pass = 2
          DOUBLE PRECISION, PARAMETER :: dl = 5D-3
          DOUBLE PRECISION, PARAMETER :: stepsize(3)=(/0.25,0.05,0.01/)
 
@@ -751,7 +750,7 @@ MODULE beams3d_physics_mod
          !     Local variables
          !--------------------------------------------------------------
          LOGICAL          :: ltest
-         INTEGER          :: ier, l, m, nbeampass
+         INTEGER          :: ier, l, m
          DOUBLE PRECISION :: rinv, phi_temp, dt_local, ti_temp, ne_temp,&
                              s_temp, x0, y0, z0, xw, yw, zw, te_temp, Zeff_temp, &
                              rho_temp, rlim, zlim,vtor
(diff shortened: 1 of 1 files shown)


kudav (commented):
The deposition works as expected, changing the allocation between deposited/port/shinethrough particles slightly for the tangential beams at high energy. However, I would propose fixing the bug by moving the 'end do' statement of the multi-pass loop to the end of the function, before the return statement. From what I can tell, this recovers the desired behavior:
grafik
Left: old multipass, middle: this PR, right: old multipass with the proposed fix

lazersos:
@kudav if you appove the merge, we can base a new pull request with your proposed modification for testing.

kudav (approved):
Fine by me

cfe316:
Thanks for addressing this issue! It looks like the middle image still has one particle from the yellow Q6/8 beam and about 5 particles from the blue Q3 beam which have escaped. So there's still something funny happening.

kudav:
I agree, and I dont want to rule out a bug with the wall collision algorithm, however note in the figure above that the red contours mark the plasma and not the wall, which in this case is a detailed 3D wall and not drawn. I think singular particles can pass through the ports (which are not closed) opposite to the NBI. 
I do also remember seeing particles passing through walls for more simple wall structures in other runs I did.

PR #349, opened 2025-02-16
Feature/sopt bootprox

lazersos:
This pull request adds the B(m=1,n=0)/B(m=1,n=1) target to STELLOPT. This target is a proxy for the bootstrap current. The proxy is based on findings in the development of W7-X where B10/B11~0.5 minimized the bootstrap current [Phys. Fluids B 5, 3728–3736 (1993)](http://aip.scitation.org/doi/10.1063/1.860843). It should be noted that there is also an orthagonal dependence on B(m=0,n=1) which controls the mirror term. The user is free to specify the target B10/B11 ratio.


lazersos (commented):
Shouldn't interfere with others works.

PR #351, opened 2025-02-17
feature/dkes_printout

ajchcoelho:
Variable `lprint` which was already in place, now is finally used. It is no longer an OPTIONAL variable since Fortran 77 does not support it. `lprint` is useful when we don't want DKES output files to be written (when using THRIFT for instance).

In the future, an improved version of this is to separate the printout routine in 2: one that does the calculations of `Lij` and another that actually writes stuff to file.


lazersos (approved):
Looks reasonable.

PR #356, opened 2025-02-25
fmin_acceptable value change

### LINKED ISSUE #354: BEAMS3D/FIELDLINES: Bug (and solution) for failures of inverse lookup.
Running BEAMS3D with AUG equilbria from VMEC (with ns=512, ftol_end=1e-19) produces wrong fields on some individual grid points. This extends to the BR, BZ and U values, but sometimes also to BPHI and S. Ive tried both with and without -plasma with essentially the same results. Changing the grid resolution changes where these grid points occur.
The black dots in the left panel show the R and Z locations of the VMEC grid, the right panel the differnce of the U_ARR variable in the R direction to clearly show the 0 values. Here, nr=301, nz=501.
Image
There were some changes to the functions in vmec_utils in the last year or so, which get perform the inverse lookups to construct the cylindrical grid. Is it possible that these introduced bugs?

lazersos:
This pull request closes #354 by setting the `fmin_acceptable` value to `1.0E-8_dp`.

### DIFF
--- a/LIBSTELL/Sources/Modules/vmec_utils.f
+++ b/LIBSTELL/Sources/Modules/vmec_utils.f
@@ -53,7 +53,7 @@ C-----------------------------------------------
 C-----------------------------------------------
 C   L o c a l   V a r i a b l e s
 C-----------------------------------------------
-      REAL(rprec), PARAMETER :: fmin_acceptable = 1.E-12_dp
+      REAL(rprec), PARAMETER :: fmin_acceptable = 1.E-8_dp
       INTEGER     :: nfe, info_loc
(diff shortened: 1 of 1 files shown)


kudav:
Looks good. From my testing, nothing speaks against going down with the tolerance to 1e-9, which is what I suggest we do. An alternative could be to just set this as standard value and be able to control the parameter via the namelist.

lazersos:
Problem is multiple codes make use of this library. For now let's leave it at `1E-8`.

kudav (approved):
Alright, looks good.

PR #360, opened 2025-03-10
Feature/magmat sections

lvanham:
This branch contains MUMAT (the magnetostatics code) as well as modifications to BEAMS3D that allow it to make use of the former code.


lazersos (approved):
Looks pretty good to me overall. @kudav please run an ASDEX and W7-X deposition case to verify that they run correctly.

kudav:
Both AUG and W7-X deposition runs are normal, as in the develop branch (checking the output file and comparing the deposition information given there).

PR #362, opened 2025-03-26
LIBSTEL: Fixed header issue in Map files regarding large int.

lazersos:
Fixes an issue where the `MAP` type optimizations had a bad integer write for very large integers.


ajchcoelho (approved):
Looks good.

PR #363, opened 2025-04-03
Merge of pressure and density solvers into develop

ajchcoelho:
1. The main marge in this pull request is to bring-in the _pressure_solver_ and _plasma_solver_ classes into **pySTEL**. 
2. It further makes modifications in **PENTA** code in order to compute the neoclassical coefficients which are used by the solvers. 
3. Modifications are made to other **pySTEL** classes (penta.py, collisions.py, dkes.py, fusion.py, popcon.py and plasma.py)
4. Minor modifications regarding printing and output are made to the **THRIFT** code.
5. Small modifications to **DKES** code. These are the exact same commits as in an [already closed pull request](https://github.com/PrincetonUniversity/STELLOPT/pull/351).

An example case on how to use the pressure and plasma solvers will be provided later.


lazersos (approved):
Looks good. At some point we probably want to wrapper the solver routines with a _util routine to aide in running the code.

PR #366, opened 2025-04-25
BEAMS3D: Changed BEAMS3D to handle 6 background ion species

lazersos:
As the title says, it's a modification so BEAMS3D can handle 6 background ion species.


kudav (approved):
Looks good to me and compiles as before.

PR #367, opened 2025-04-26
Bugfix for BEAMS3D VMEC field lookup

lazersos:
These changes help improve the interface between BEAMS3D and VMEC. BEAMS3D requires mangetic fields in cylindrical coordinates on a cylindrical grid. VMEC stores quantities on an `s,u,v` grid where `v` is the toroidal angle normalized to 2pi. Thus a 2D inverse lookup is used to find a given set of `s,u` coordaintes for a desired position in `R,Z` at a given toroidal angle. To do this VMEC (`vmec_utils.f`) uses a 2D newton method. This lookup sometimes fails becasue a certain tolerance was reached. Previously when this happend the `s,u` coordiantes were returned but no magnetic field data. This could lead to 'holes' in the magnetic field. To rectify this we now use a multi-pass method  in `beams3d_init_vmec.f90` to try to improve the inital `s,u` guess. If no appropriate lookup can be found then we average in R,Z the neighboring points.  In the future we should attempt to improve the 2D newton method but for now these changes seem to fix things.

This has been tested against the BENCHMARK problems and against a W7-X case.

### DIFF
--- a/BEAMS3D/Sources/beams3d_init_fusion.f90
+++ b/BEAMS3D/Sources/beams3d_init_fusion.f90
@@ -84,7 +84,7 @@
       CALL init_random_seed
       IF (lverb) THEN
          WRITE(6, '(A)') '----- INITIALIZING FUSION REACTIONS -----'
-         WRITE(6, '(A,I6)') '      nparticles_start: ', nparticles_start
+         WRITE(6, '(A,I8)') '      nparticles_start: ', nparticles_start
          CALL FLUSH(6)
       END IF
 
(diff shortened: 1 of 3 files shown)


kudav (approved):
Looks good, this seems to fix the remaining issues present with using AUG VMEC equilibria. Below is the second derivative of BR (in R direction) for the interface before these fixes (left), the first fix (middle) and the latest commit (right). The erroneous grid point at R=1.5, z=0 is removed in both fixed cases. Some additional oscillation appears for the latest commits around the outboard midplane, otherwise both appear identical.
{0929DB93-7D39-442F-AF7D-27A00906B4DE}

PR #368, opened 2025-05-02
Feature/root analysis

ajchcoelho:
I am changing the criterium `THRIFT` uses to pick one of the ambipolar roots from `PENTA`. Up to now we were using Maxwell's construction criterium. The problem is that I realized this method is very sensible to resonances that sometimes appear in the fluxes (these resonances are inherent to `DKES` formulation and there is nothing we can do about them).

So now, whenever there are 3 ambipolar roots, we simply pick the first one (lowest Er).
If number of roots is 5, we pick the negative root closest to Er=0, according to the following 2 possibilities:

`! possibility_1 = ['extra_stable_root','unstable_root2','ion_root','unstable_root','electron_root']`
` ! possibility_2 = ['ion_root','unstable_root','electron_root','unstable_root2','extra_stabe_root']`

If number of roots is even or larger than 5, an error is given and the code STOPS -- as was already the case.

Finally, in this merge I am changing the splines of `Er`, `J_BS` and `eta` to Hermite, to prevent spurious oscillations.

### DIFF
--- a/PENTA/Sources/penta_interface_mod.f90
+++ b/PENTA/Sources/penta_interface_mod.f90
@@ -1426,23 +1426,23 @@ MODULE PENTA_INTERFACE_MOD
    END SUBROUTINE penta_merge_fluxes_vs_Er_files
 
    SUBROUTINE root_analysis
-      ! using the Maxwell construction criterium, whenever num_roots>1, determines which of them will settle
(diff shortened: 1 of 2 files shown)


lazersos (approved):
Seems reasonable to me.

PR #372, opened 2025-05-20
Bugfix/beams3 d ascot5

lazersos:
Updates for the latest version of ASCOT5.  Also fixes possible bug in interface.


kudav (approved):
AUG case ran fine with these changes. Did not check if ASCOT would run with the outputfile though as I dont have it currently installed. Note that we could improve the information passed to console regarding the writing of the ASCOT outputfile.

PR #375, opened 2025-06-02
Feature/thrift_travis

ajchcoelho:
Calls to `TRAVIS` from `THRIFT` should be made using sqrt of normalized flux and not the normalized flux. This bug is fixed. 

The logic on how to ramp-up/down ECRH power using the NAMELIST variables `PECRH_AUX_T` and `PECRH_AUX_F` is improved.


lazersos (approved):
Looks good to me.

PR #377, opened 2025-06-16
Feature/coil optimization

lazersos:
This pull request adds the ability to do free boundary coil optimization in STELLOPT. The coil is not parameterized by splines in rho, theta, and zeta. Here rho is in units of meters. Targets include bnormal on the boundary of the plasma, coil torsion, coil curvature, and coil-coil spacing. The interfaces to COILOPT++ and REGCOIL have been depricated in favor of this new capability. Future improvements will include brining back the ability to generate an MGRID file, thus allowing for coil-plasma optimization, and possibly REGCOIL.


ajchcoelho:
OK to me!

PR #381, opened 2025-07-24
Change THRIFT edge boundary condition and improve others

ajchcoelho:
The main changes of this pull request are:

1) Edge boundary condition of `THRIFT` is changed to include a pprime term (diamagnetic) which was missing and is important to ensure that the simulation reaches the expected steady-state. This new BC follows equations (D8) and (D13) of Schmitt PhD Thesis

2) Spline of parallel resistivity coming from PENTA is made with the log. This improves the spline as etapar spans over several orders of magnitude

3) EparallelB is now correctly computed using Eqs. (3.61) , (3.64) and (3.65) of Schmitt thesis. Once again, the previous formula was relying on `THRIFT_JPLASMA` thus missing a pprime term

4) Improves `vmec.py` and `thrift.py`


lazersos (approved):
Looks good to me.

PR #383, opened 2025-08-14
Inclusion of the LgradB metric in STELLOPT

lazersos:
This pull request adds support for the L_grad(B) coil metric in STELLOPT. A subroutine for computing L_grad(B) has been added to `stel_tools` and checked against the results from ["The magnetic gradient scale length explains why certain plasmas require close external magnetic coils"](https://doi.org/10.1088/1361-6587/ad1a3e).  Additionally, a bug was discovered in how the edge half-grid quantities were being extrapolated to the full grid in `stellopt_load_equil`.


ajchcoelho:
Looks good.

ajchcoelho (approved):
I believe it's good to merge.

PR #391, opened 2025-10-08
bugfix/vmec vseq

lazersos:
This patches VMEC so the `VSEQ` namelist works again.  Adds variable explanation to namelist.


ajchcoelho (commented):
Looks OK to me.

lazersos:
I still need to check that this doesn't crash STELLOPT, but I don't think it should. If anything it 'could' improve stability a bit.

PR #392, opened 2025-10-09
Feature/plasma solver updated

ajchcoelho:
This merge is mostly about bringing a plasma solver (particle and energy transport equations) into `THRIFT`. Before merging, there are two essential things that need to be checked:
1. `THRIFT` plasma solver is doing the same thing as the plasma solver of `pySTEL`;
3. The evolution of current/iota by `THRIFT` is the same, whether it uses the profiles computed _a priori_ by `pySTEL`, whether it solves the transport equations simultaneously to the current evolution.

The first point is verified by running `THRIFT`  between t=0.0 and t=29s, with `dt_THRIFT=29s` and `dt_plasma_solver=0.01s`. This means means that THRIFT is not evolving current; only transport quantities are evolved. The input file for this test is
[input.GIGA_v515.txt](https://github.com/user-attachments/files/22792045/input.GIGA_v515.txt). Note that we are not interested in the evolution of current, we really just want to check the plasma solver. Because current is not evolving, this means that `aminor`, `vprime`, `phi_edge` and `B_squared` (which are parameters that enter into the transport solver) remain constant throughout the whole simulation and are those that `VMEC` computes at t=0.0s. The comparison of temperatures and densities at t=29s shows that both codes are evolving the transport equations similarly:

image
image

This thus confirms that the integration of the plasma solver in THRIFT was successful, with both `pySTEL` and `THRIFT` transport solvers doing the same thing.

The second point is first verified by running the current version of `THRIFT` in `develop` (we call it 'old') and the 'new' version of `THRIFT` (i.e. the one we are about to merge), where in both simulations the plasma profiles are provided from solving the transport equations _a priori_ in `pySTEL`. We have checked and the comparison is 100% exact, meaning that the modifications brought to the code by this merge DO NOT affect the current evolution.

Finally, we compare the simulation in the 'old' branch where profiles are provided with a simulation in the 'new' branch where the transport equations are also evolved by `THRIFT`. Now, this comparison is a bit more subtle since when the current is evolving then `aminor`, `vprime`, `phi_edge` and `B_squared` are also evolving in time, and as a consequence the plasma profiles will differ a bit:

image

(As an aside node, this actually highlights the importance of solving the transport equations in a self-consistent way by coupling with an equilibrium solver! ) When we compare "old" vs "new" `THRIFT` quantities, we see that these differences mostly affects the bootstrap current (as expected, since JBS has a strong dependency on the profiles):

image
image
image
image

We underscore that these differences between 'old' and 'new' are simply a result of the way the comparison is conducted. The only way to have an exact match would be to either also evolve the aforementioned parameters in `pySTEL` (which is virtually impossible since this basically would mean having `VMEC` and `THRIFT` implemented in `pySTEL`) or to fix these parameters inside `THRIFT` plasma solver to the values computed by `VMEC` at t=0.0s. Such test, I believe, is useless and redundant since we have already clearly demonstrated that:
1. `pySTEL` and `THRIFT` plasma solvers are identical;
2. This merge does not affect at all the current evolution parts of the code.

Other minor changes to some `pySTEL` classes are also included in this merge.


lazersos (approved):
Looks good to me for mergeing.

PR #401, opened 2025-11-10
Feature/plasma solver faster

ajchcoelho:
This merge concerns the `pySTEL` transport solver embedded in the plasma_solver class. The solver has been highly optimized (it is 7x faster in some runs) and is now properly commented.


lazersos (approved):
Looks good to me.

PR #408, opened 2025-12-01
New features to THRIFT and to pySTEL/vmec

ajchcoelho:
This merge brings the following new features to `THRIFT`:
- we can now provide non-constant heat diffusivities via the input h5 file. This allows for more realistic transport simulations;
- the calls to `VMEC` are done using as input `dI/ds` computed directly from `THRIFT_I` and not through `THRIFT_J`. This makes the iota profiles smoother in some situations.

The merge brings some novelties to `pySTEL` too, the most important ones being on `vmec.py`: 
-  a routine that calculates the covariant tensor `g_ij` has been created;
- the routine that gives `<|nabla_rho|^2 >` has been fixed;
- a routine that computes `<|nabla_rho|>` has been added.


pySTEL/libstell/vmec.py
```diff
@@ -199,58 +199,214 @@ def calc_magwell(self):
...
+		if self.iasym==1:
+			R = R + self.sfunct(theta,phi,self.rmns,self.xm,self.xn)
+			Z = Z + self.cfunct(theta,phi,self.zmnc,self.xm,self.xn)
+		#
+		s = np.linspace(0,1,self.ns)
+		#
+		dRds = np.gradient(R,s, axis=0)
+		dZds = np.gradient(Z,s, axis=0)
+		#
+		dRdu = np.gradient(R, theta.ravel(), axis=1)
```
lazersos: dX/du and dV/du should be calculated from the derivatives of the Fourier Harmonics for a more accurate representation.

lazersos (changes requested):
There's a minor point here about using the derivatives of the functions and not `gradient` to compute dX/du and dX/dv.

pySTEL/libstell/vmec.py
ajchcoelho: Done. Now the poloidal and toroidal derivatives are computed via the Fourier Harmonics and not through np.gradient. 

I have also changed the way the radial derivatives, d/ds, are performed, by mimicking what vmec does internally. Now even m-modes and odd m-modes are treated differently, the former interpolated as `s` and the latter as `sqrt(s)`. This overcomes the issue of the derivatives diverging close to the axis.

An interesting test to the newly implemented `calc_covariant_tensor` function in `vmec.py`  is to compare the jacobian of an wout file with that computed through the determinant of the g_cov matrix returned by  `calc_covariant_tensor` . They agree:

image

ajchcoelho:
I have also introduced a bugfix to THRIFT. When PHI_EDGE<0, `THRIFT` was not caring about its sign, which it turns out to be quite important.

lazersos (approved):
Looks good now.

PR #421, opened 2026-01-21
SHARE: added makefile for Niflheim DTU

luisahuebner:
This is the new makefile for running on the supercomputer Niflheim at DTU.


lazersos (approved):
Looks good to me.

PR #422, opened 2026-01-23
Modifications to the transport solver PID's and changing of gyro-Bohm definition

ajchcoelho:
With this merge:
- I have modified, for clarity, some of the variables names of the PIDs. For instance, the `total_power` is now the `max_total_power` and the `time_dependent_factor` is now `time_dependent_electron_temp_axis`, etc
- I corrected the definition of Q_gB (used to denormalize the turbulence fluxes) to be in accordance with the definition in `stella`, from which the normalized fluxes usually come from
- the `plasma_solver_util.py` has been cleaned and slightly improved

An example of a script that launches a transport simulation where density is controlled via a PID can be found [here](https://github.com/user-attachments/files/24825419/PID_with_neon.py)

To think for the future: dome functions inside the classes `plasma_solver`, `thrift`, etc, are more of "auxiliary mathematical functions" than actually essential parts of the classes. These are for instance: `akima_derivative`, `polyfit_derivative`, etc. I wonder if at some point it would make sense to have an `auxiliary_functions.py` file inside `pySTEL/libstell`.

### DIFF
--- a/LIBSTELL/Sources/Modules/thrift_globals.f90
+++ b/LIBSTELL/Sources/Modules/thrift_globals.f90
@@ -55,7 +55,7 @@
       INTEGER, PARAMETER :: nions_max = 6
       REAL(rprec) :: dt_plasma_solver, tol_plasma_solver
       REAL(rprec) :: stiffness_beurskens, aLT_critical_beurskens, alpha_beurskens, &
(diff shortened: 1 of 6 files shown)


lazersos (approved):
Looks good to me.

PR #426, opened 2026-02-02
FIELDLINES: restart bug fixes

pponsv:
This pull request tries to fix two bugs in FIELDLINES, both occurring when restarting calculations from another run. 

- The first solves a deallocation error when using MPI and hdf5. The array allocation in the regular run was done with `mpialloc`, while when using the restart flag the arrays were allocated per process and the data was then shared using `MPI_BCAST`. When calling `fieldlines_free`, `mpidealloc` was called on `B_R, B_PHI, B_Z, raxis, phiaxis` and `zaxis`, and an error was raised. The solution is to initialize these variables with `mpialloc`. 

- The second solves another issue (that I only encountered during -restart runs, but that could be more general) with the spline interpolation of the diffusion coefficient. In line 123 of `out_fieldlines_nag.f90`, the positions are updated as
`q(1) = q(1) + random_normal() * sqrt(fval(1)*ABS(dphi))`
Sometimes, the spline interpolation overshoots (maybe related to #354 ?) and `fval` is negative, with the `sqrt` returning a `NaN`. The proposed fix for this is to use linear interpolation, adding `MU_spl%isLinear=1`  to the spline initialization.

### DIFF
--- a/FIELDLINES/Sources/fieldlines_init.f90
+++ b/FIELDLINES/Sources/fieldlines_init.f90
@@ -268,6 +268,7 @@
             CALL EZspline_init(MODB_spl,nr,nphi,nz,bcs1,bcs2,bcs3,ier)
             IF (ier /=0) CALL handle_err(EZSPLINE_ERR,'fieldlines_init:MODB_spl',ier)
             MODB_spl%isHermite = 1
+            MU_spl%isLinear=1
             MODB_spl%x1 = raxis
             MODB_spl%x2 = phiaxis
             MODB_spl%x3 = zaxis
--- a/FIELDLINES/Sources/fieldlines_init_restart.f90
+++ b/FIELDLINES/Sources/fieldlines_init_restart.f90
@@ -40,42 +40,45 @@ SUBROUTINE fieldlines_init_restart
       CALL read_scalar_hdf5(fid,'nr',ier,INTVAR=nr)
       CALL read_scalar_hdf5(fid,'nphi',ier,INTVAR=nphi)
       CALL read_scalar_hdf5(fid,'nz',ier,INTVAR=nz)
-      ALLOCATE(raxis(nr),phiaxis(nphi),zaxis(nz))
+      CALL close_hdf5(fid,ier)
+   end if
+#endif
+#if defined(MPI_OPT)
+   CALL MPI_BCAST(nr,   1, MPI_INTEGER, master, MPI_COMM_FIELDLINES, ierr_mpi)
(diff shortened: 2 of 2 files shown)


FIELDLINES/Sources/fieldlines_init.f90
```diff
@@ -268,6 +268,7 @@ SUBROUTINE fieldlines_init
             CALL EZspline_init(MODB_spl,nr,nphi,nz,bcs1,bcs2,bcs3,ier)
             IF (ier /=0) CALL handle_err(EZSPLINE_ERR,'fieldlines_init:MODB_spl',ier)
             MODB_spl%isHermite = 1
+            MU_spl%isLinear=1
```
lazersos, line 271: This is dangerous and will break the code with certain compilers.  The code is hardcoded to use `R8HERM3FCN` to evaluate splines. Both `EZspline_init` and `Ezspline_setup` cannot guaruntee that the un-used parts of `MU_spl%fspl` are set to zero. The correct change is to set `MU_spl%isHermite = 1`.  Anything else would possibly break the code. @pponsv, please test this and see if the oscillation still exists.

lazersos (changes requested):
Current use of isLinear will break code on some compilers.

FIELDLINES/Sources/fieldlines_init.f90
pponsv: MU_spl%isHermite = 1 is already set in line 298 (I made a mistake when preparing the pull request and the changed line should have been after that), so that won't fix the issue. Is there no way to use linear interpolation then?

lazersos: Not without significant modification to existing routines. Can you provide and example of this oscillation that you find in MU?  I suspect the issue is that you're using a VMEC equilibrium and the oscillation is not due to MU persay but the fact we multiply by the field to get account for pushing in phi not real space 
```fortran
MU3D = ABS(2*mu*sqrt(B_R**2+B_PHI**2+B_Z**2)/B_PHI)
DO i = 1, nr
   MU3D(i,:,:) = raxis(i) * MU3D(i,:,:)
END DO
```
The answer is that such oscillations associated with your equilibirum not FIELDLINES. This is also why you can't use linear interpolation.  Best suggestion is to only use diffusion with vacuum fields or a type of equilibrium which handles the edge (like HINT).

lazersos:
@pponsv Also for various reasons we do not allow pull requests from forked repositories. If you wish to commit changes to this repository please create a properly titled `feature/name` or `bugfix/name` branch in this repository with the appropriate changes.

PR #428, opened 2026-02-06
Fix DKES_D31 sign and implement diffusion+advection in transport solver

ajchcoelho:
With this merge:
- We fix the sign of DKES_D31. After comparing `JBS\cdot B>` with a W7-X discharge and with a tokamak case, I am now certain that this is the correct way to use this coefficient;
- The `pySTEL` transport solver can now run with rho- and a/Ln-dependent particle diffusion and advection coefficients. This allows to use data from `stella` to model the particle transport.


lazersos (approved):
Looks fine to me.

PR #429, opened 2026-02-12
Total coil twist as optimization target

MaxRang:
Implemented total coil twist as an optimization target via the integral of the torsion


lazersos:
@MaxRang please do not merge in forked repositories. Our policy is for changes to this repository to be directly tracked here.  If you want to work on a forked repository that's fine, but pull requests should be using branches in the PrincetonUniversity repository.

PR #431, opened 2026-02-15
Added new coil chisq targets to STELLOPT

lazersos:
Added the integral total torsion and the coil distortion metrics to STELLOPT. Also fixed an issue where the FIELDLINES namelist was improperly output which could crash the code.


LIBSTELL/Sources/STELLOPT/stellopt_input_mod.f90
```diff
@@ -961,8 +962,8 @@ SUBROUTINE init_stellopt_input(lfull_reset)
       sigma_bnmnc              = bigno
       target_coil_curvature    = 0.0
       sigma_coil_curvature     = bigno
-      target_coil_torsion      = 0.0
```
MaxRang, line 964: Mean coil torsion should still exist as a target, so default values should still be set, no?

lazersos: @MaxRang Mean_coil_torsion hasn't been merged into develop yet.  Will add you as a developer so you can issue the pull request from a branch not a fork.

loredanamihaelavasiloiu (approved):

PR #432, opened 2026-02-23
Added capability to generate STELLOPT coil representation from coils file.

lazersos:
This pull request fundamentally allows the user to generate the `RHO_COIL_KTS`, `THETA_COIL_KTS`, and `ZETA_COIL_KTS` STELLOPT arrays from a coils file. To achieve this, parts of `coil_spline_mod.f90` were reafactored to define routines which go from the rho, theta, zeta representation to x,y,z. Additionally the opposite transformation was possible by using that method and a 2D newton method. Finally, instead of constructing splines over rho and theta to create the coils, an `Rx=cos(theta)` and `Ry=sin(theta)` coordiante set is now used. This avoids the requirement that theta knots explicitly need to be at `theta=0`. 

There was also a fix for some missing variables when reading in a wout file and trying to write it as an `INDATA` name list.


MaxRang (approved):
Looks good to me. Ideally, one would be able to construct the rho, theta, zeta splines from the coils file alone, rather than needing a VMEC object, but I realise this is not easy

lazersos:
> Looks good to me. Ideally, one would be able to construct the rho, theta, zeta splines from the coils file alone, rather than needing a VMEC object, but I realise this is not easy

Well without the VMEC equilibrium the notion of `RHO` and `THETA` are not defined. However, the way the subroutines are written, any Fourier decomposed boundary can be used.

PR #435, opened 2026-03-02
Update fieldlines_input_mod.f90

MaxRang:
There was a field missing in the fieldlines_input namelist, causing the following error:

Invalid line in namelist:   MUMATERIAL_NNEIGHBOR = 100
ERROR reading namelist FIELDLINES_INPUT from file:

The field was written to the input file, but did not exist in the namelist. Adding the mumaterial_nneighbor variable to the namelist should solve the issue.


lazersos:
Error was fixed in a previous pull request. MUMATERIAL_NNEIGHBOR should no longer be written out the namelist.

lazersos (approved):

PR #442, opened 2026-03-23
THRIFT Improvements

ajchcoelho:
This merge brings the following major changes to `THRIFT`:
- Fixes `thrift_bootsj` (the sign of JBS.B was not being considered!)
- `thrift_penta` is now faster. Before, when the plasma solver was ON and add_NEO=True, the electric field was computed every dt_plasma, which would make simulations with NEO fluxes very slow and infeasible. Now, a new ambipolar solution is only computed at every dt_THRIFT
- `penta_interface_mod` is now more robust, as it allows any number of odd Er solutions (before an error was given if number of solution was larger than 5)

`pySTEL`  has also been upgraded, namely the `thrift`, `dkes` and `plasma` classes.


lazersos (approved):
looks good

PR #445, opened 2026-04-02
PENTA interfaced to STELLOPT

lazersos:
This pull request interfaces the PENTA code to STELLOPT. This required a bit of refactoring when it came to the DKES interface. In particular we've introduced an `LNEED_DKES` and `LNEED_PENTA` arrays internally to help with sorting of data. Because of this change the `TARGET_DKES_ALPHA` and `TARGET_DKES_ERDIFF` targets need to be redone and for now are depricated. PENTA based targets include `TARGET_PENTA_ER` and `TARGET_PENTA_J`. Also the DKES NU/ER pairs are now only written once to the `OPTIMUM` namelist.


lazersos (commented):
Looks good. We should check in the future if we can use the derivative terms elsewhere in the code (where we currently do finite difference). Also I'm not sure power_series_rho and power_series_rho2 are entirely correct. Becasue I think what we want this function to return is always df/ds not df/dx where x is the parameterization.  Otherwise good.

ajchcoelho:
I fixed the derivative of `power_series_rho` and `power_series_rho2`. Besides `stellopt_penta`, I believe the only part of the code where derivatives of the profiles are taken is in `stellopt_sfincs`. Correct?

lazersos:
I believe that is correct.

PR #446, opened 2026-04-08
Feature/libstell transport

ajchcoelho:
This creates a module for transport in `LIBSTELL`. The purpose is to have inside this module routines that are called from `libstell.py` in order to make the `pySTEL` transport solver (and others)  faster.


lazersos (approved):
Looks good to me.

PR #447, opened 2026-04-14
Prevent unnecesary stellopt_load_equil() calls

MaxRang:
During coil optimization, the equilibrium does not change, hence we don't need to reload it every iteration

### DIFF
--- a/STELLOPTV2/Sources/General/stellopt_fcn.f90
+++ b/STELLOPTV2/Sources/General/stellopt_fcn.f90
@@ -48,7 +48,7 @@
 !        iflag       Error flag
 !        iunit       File unit number
 !----------------------------------------------------------------------
-      LOGICAL ::  lscreen
+      LOGICAL ::  lscreen, lload_equil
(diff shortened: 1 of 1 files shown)


loredanamihaelavasiloiu (approved):

lazersos:
This is a pointless modification. It does nothing.

MaxRang:
Could you elaborate? If I made an error, I would like to understand.

My intention was for the modification to ensure that stellopt_load_equil() does not run on calls of stellopt_fcn() where iflag .ne. -1. This would mean that stellopt_load_equil() is skipped whenever VMEC is not called, which is every iteration in the coil optimization.

lazersos:
@MaxRang I see what you're doing but this is complicating the logic quite a bit. I think this is a fine thing for you and @loredanamihaelavasiloiu to implement in your local branches but I'm very hesitant to implement it for everyone.

MaxRang:
That's fair. I agree that the solution I came up with is far from elegant.

PR #451, opened 2026-04-16
FIELDLINES: Implemented scaling factor for mumaterial B-fields.

lazersos:
This pull request adds a scalar to the FIELDLINES_INPUT namelist called `mumaterial_scale` which linearly scales the B-field calculated from the magnetic material. This is done to mimic the effect of the material having less material. The default value is `1.0` so that not setting this values has the intended behavior.


luisahuebner (approved):
looks good.

PR #456, opened 2026-05-08
Free Boundary Coil Optimization

lazersos:
This adds an ability to STELLOPT to perform free boundary coil optimization. This is sometimes called single stage optimization. To run in this mode, not only must coil optimization be performed but the VMEC `INDATA` namelist will require `LFREEB=T` and the user to provide an inital `MGRID_FILE` which VMEC can use to run. This MGRID file will only be read once and STELLOPT will generate new MGRID files as it runs. This is attributed to the fact that when VMEC reads the `INDATA` namelist it automatically checks the `MGRID_FILE` and loads it. If it does not load, the code will set `LFREEB=F` and then STELLOPT will behave as if fixed boundary optimization is happening.

Please check to make sure non-coil optimization runs are not affected.


BENCHMARKS/STELLOPT_TEST/COIL_OPT/input.COIL_OPT
```diff
@@ -216,30 +214,74 @@
   NFUNC_MAX   = 5
   OPT_TYPE    = 'GADE'
   EQUIL_TYPE  = 'VMEC2000_ONEEQ'
```
viper2642, line 216: perhaps delete overloaded definition of EQUIL_TYPE

viper2642 (commented):
a very minor suggestion to delete overloading definition in input. Other than that, it feels ok.

PR #458, opened 2026-05-18
Feature/parallelize booz xform

kudav:
This PR parallelizes the booz xform code over flux surfaces. Speed scales with processor count as expected.


BOOZ_XFORM/Sources/boozer_coords.f
lazersos, line 1: This breaks the STELLOPT interface.

BOOZ_XFORM/Sources/boozer_setup.f
lazersos, line 1: This breaks the STELLOPT interface.

lazersos (dismissed):
The proposed modification breaks the STELLOPT interface to BOOZ_XFORM.

BOOZ_XFORM/Sources/boozer_setup.f
kudav: I pushed a fix for the interfaces to STELLOPT, and THRIFT, which used the same lines of code. Unfortunately, my compilation of STELLOPT does not work currently (there is a problem with NESCOIL), so I cannot check if the fix is successful.

lazersos:
This still doesn't work for STELLOPT. When you've got it working for STELLOPT and THRIFT and can test then I think the pull request should be made. Right now this change adds minimal utility and breaks other codes.

kudav:
So after jumping through some hoops I managed to get thrift_boozer to be called with a modified version of the included test case, and the boozer conversion ran fine - both with 1 and 16 processors. I did not manage to run any of the stellopt test cases, on the MPCDF clusters stellopt got stuck compiling PENTA and on tok I stellopt crashed after initialization in MPI_Comm_free in all test cases I tried.
Could you specify what exactly you meant by 

> This breaks the STELLOPT interface.

lazersos:
What is the PENTA compilation error?

Also retesting now.

lazersos (commented):
Code looks OK now....still need to verify with test cases.

lazersos (approved):
Looks good now.

PR #459, opened 2026-05-18
VMEC_UTILS: improve inverse lookup stability

kudav:
This implements a few more fixes to the inverse lookup for the VMEC magnetic field. Initial testing was very positive in finding edge cases that were not addressed before but I'd like to document this more before merging.

EDIT: some more documentation on the changes this implements, and verification.

- If the residual fmin is already below fmin_acceptable, GetBcyl_WOUT now returns a success even when info_loc=-1 from cyl2flux. Note that this guard only makes the implementation consistent with the already existing ones in other GetBcyl functions (line 208, 350, 516). If there is a reason why it should not exist in this function, we can revert.
- Inceased the number of possible retries in cyl2flx
- Bugfix in newt2d: the gradient used to be computed using the information of the previously rejected point, instead of the information in xc_min, which should be reverted to at this point and ensures the point and gradient are consistent.
- Smoother damping: instead of a threshold, the damping factor is now continuous and shrinks proportionally to the residual.
- The tolerance for degenerate Jacobian case has been decreased and an attempt to find a valid point by jogging the angle and pushing away from the axis is made.
- In fieldlines, I moved sflx=0 to before the loop starts so that the information is carried to the next iteration. Thinking more about it this is likely more trouble than its worth due to wrap-around and points outside the plasma, which would need additional handling. I switched back to the previous version in the last commit.


lazersos:
While this looks rather simple, it could have profound implications for other peoples works. Could you perform a test were you compare the returned values by the function before and after the modificaiton.  R/Z at phi=0,1/4,1/2 field periods for NCSX for example.

LIBSTELL/Sources/Modules/vmec_utils.f
```diff
@@ -95,6 +95,8 @@ SUBROUTINE GetBcyl_WOUT(R1, Phi, Z1, Br, Bphi, Bz,
 !     If info == -1 then the tollerance was not achieved
 !     If info < -1 then most likely the point is outside the eq.
 !
+      IF (info_loc.eq.-1 .and. (fmin .le. fmin_acceptable)) info_loc = 0
```
lazersos, line 140: This fundamentally changes the behavior of the algorithm.

lazersos (commented):
I think we need to see that BEAMS3D for example produces consistent results.

lazersos:
I tested with BEAMS3D and FIELDLINES for our test cases. Seems like results don't change. But I think it should still be documented better what this change fixes. Also the point keeping the in-code comment consistent with the behavior should also added.

kudav:
Here is the before & after for an AUG equilibrium. I'm still investigating the one gridpoint on the inboard midplane which appeared using the updates.
{6BF1660E-4DEF-498A-8F7D-3FB1CC3E8F4D}

krystophny:
I tested this branch at `dfe054a0` with the same W7-X `32 x 16 x 32` grid using 1, 2, and 48 MPI ranks.

- 1 versus 2 ranks is bitwise identical for `S_ARR`, `U_ARR`, all three field arrays, `S_lines`, `B_lines`, and the default mask.
- 1 versus 48 ranks retains the same default mask. Nineteen `S_ARR` cells differ, with a `1.01e-8` maximum. The largest field-component difference is `1.96e-8 T`; marker `S_lines` and `B_lines` are bitwise identical.
- Relative to current `develop` at 48 ranks, the default mask is unchanged and the largest marker-field difference is `2.97e-9 T`.

The remaining rank dependence comes from the caller's rank-local warm-start path, not the inverse-solver changes here. #471 removes that caller state and the shared in-place repair race. With #471, 1, 2, and 48 ranks are bitwise identical for all listed arrays and masks.

pponsv:
In `cyl2flx`, during the loop, if newt2d has not converged by the end the angle is rotated anyway, so the returned poloidal angle is pretty far away from the (unconverged) one, even if `fmin` is pretty small.

```
DO itry = 1, 8

   CALL newt2d(xc_opt, fmin, ftol, nfe_out, nvar, info)
   nfe = nfe + nfe_out

   IF (fmin.le.ftol .or. info.eq.-3) EXIT
!
!        JOG POINT (BY ROTATING ANGLE) TO IMPROVE CONVERGENCE
!
   xc_opt(2) = xc_opt(2) + twopi/8
   IF (MOD(itry,2).eq.0) xc_opt(1) = one/(2*(ns_loc-1))

   fmin0 = MIN(fmin, fmin0)

END DO
```

The solution for that would be to move the angle jogging to before the call to newt2d:

```
DO itry = 1, 8
   IF (itry .gt. 1) THEN

      !  JOG POINT (BY ROTATING ANGLE) TO IMPROVE CONVERGENCE
      xc_opt(2) = xc_opt(2) + twopi/8
      !  I guess this sends the point to s~0.5 so it also belongs here?
      IF (MOD(itry,2).eq.0) xc_opt(1) = one/(2*(ns_loc-1))
      
   END IF

   CALL newt2d(xc_opt, fmin, ftol, nfe_out, nvar, info)
   nfe = nfe + nfe_out

   IF (fmin.le.ftol .or. info.eq.-3) EXIT

   fmin0 = MIN(fmin, fmin0)

END DO
```

Maybe this is what causes this weird gridpoint here?


> Here is the before & after for an AUG equilibrium. I'm still investigating the one gridpoint on the inboard midplane which appeared using the updates. {6BF1660E-4DEF-498A-8F7D-3FB1CC3E8F4D}

I have to say, with the improvements to `newt2d` I haven't found any points where this causes problems, but they do appear in the develop branch and this change fixes it.

kudav:
Thanks for the comments and suggestions. As there was some back-and-forth in this branch, I'm preparing a new PR once we have merged #478, addressing the issues and giving clear information on what this fixes.

kudav:
Superseded by #494 - closing this draft PR.

PR #460, opened 2026-05-21
THRIFT-TRAVIS integration and new THRIFT features

ajchcoelho:
I have fixed the THRIFT-TRAVIS integration and made it possible to read the power signal of each gyrotron via the thrift_profiles external file.

The merge further includes two non-TRAVIS-related developments:

1. If desired, the user can now save the current density at each THRIFT subiteration. This way we can look at why in some runs it takes many subiterations to reach convergence (or why it is not reached). This is toggled via the namelist variable `save_subiterations`
2. In `thrift_penta` we used to `CALL ROOT_ANALYSIS('ion_root')`. Now the root type is chosen via the namelist variable `Er_root_type`. The default is still `ion_root`. **Probable is a good idea to do the same in stellopt_penta??**


lazersos:
The variable `Er_root_type` is currently input via the THRIFT namelist. It would be better to include this via the PENTA `RUN_PARAMS` namelist so that the interface with STELLOPT is consistent.

lazersos:
@ajchcoelho I made the modification to move the namelist variable from THRIFT to PENTA.  I'm checking from the STELLOPT side that it works correctly. Can you verify from the THRIFT side.

lazersos:
Can verify from the STELLOPT side it works fine.

ajchcoelho:
> Can verify from the STELLOPT side it works fine.

Works well on THRIFT side.

PR #465, opened 2026-06-24
Feature/neo plasma solver

ajchcoelho:
This merge improves substantially the coupling between `PENTA`  and `THRIFT`'s transport solver. The original issue of `PENTA` is that it was not conceived to make repetitive computations of fluxes, so performance and memory were not considered important throughout the original design of the code. Here we keep intact the original `PENTA` functions and simply add new ones (dubbed `*_fast` or `*_interface`) which are then called via `penta_interface_mod`. This way, running `PENTA` via its main program still relies on the original functions. A new module for the bsplines was introduced (`bsplines_new`), copied from [here](https://github.com/jacobwilliams/bspline-fortran) and adapted/refactored to our needs. This new bsplines module is significantly faster than the original one in `PENTA`. Again, this new module is really just used when the `PENTA` functions are called from `penta_interface_mod`; when `PENTA` runs through its original main program, the old module is still used.

Since `thrift_interface_mod` changed, I also needed to slightly change `stellopt_penta`. It's really just two lines of code, but it would be good for @lazersos to make a check-run.

In order to further speed-up the `PENTA<->THRIFT` interface, I refactored `thrift_penta` and introduced a new functionality: when running the transport solver with neo fluxes, the ambipolar solution of Er is only updated at every `dt_Er_ambipolar`; this avoids having to solve the ambipolarity equation at each time-step of the plasma solver, which was one of the main reasons why the coupling `PENTA<->THRIFT` was so slow.

The pressure equation in `THRIFT`'s transport solver is now solved through `DGBSV` and no longer through `DGESV`, since the LHS matrix becomes banded after proper permutations. This leads to an impressive speed-up of the pressure solve.

A new feature was added to the transport solver: the possibility to start a simulation from some initial profiles given in the input file. This is mediated via the namelist variable `init_profiles_type`.

Finally, a bunch of improvements were made to several `pySTEL/libstell` scripts.


lazersos (approved):
Looks good to me.

PR #466, opened 2026-07-06
Bugfix/nescoil interface

lazersos:
This pull request fixes the interface to NESCOIL allowing for integration over the NESCOIL potential to produce magnetic fields inside the NESCOIL domain.


viper2642 (approved):
the interface now correctly computes the magnetic field from the surface current (potential formulation). I believe (but not tested) that the vector potential is also working

PR #467, opened 2026-07-07
Fix VMEC volume-current harmonics setup

WenyinWei:
## Summary

This PR fixes the VMEC finite-beta volume-current setup used by TORLINES and the shared VMEC volume-integral path.

The TORLINES-side fix is still the original one: the finite-beta volume-current branch now uses the VMEC Nyquist current basis consistently.

- use `mnmax_nyq`, `xm_nyq`, and `xn_nyq` for `currumnc/currvmnc`
- map `rmnc/zmns` geometry harmonics from the VMEC geometry basis onto the Nyquist basis before calling `init_volint`
- pass current arrays through temporaries whose leading dimensions match `init_volint`
- pass `MPI_COMM_FIELDLINES`, matching the existing virtual-casing path

After running the DIAGNO benchmark requested in review, I found and fixed two shared volume-integral issues as well:

- `init_volint` reconstructs the full torus (`nvp = nv*nfp`, `phi = 2*pi*xv`), so DIAGNO and TORLINES must pass the full VMEC toroidal mode number (`-xn`, `-xn_nyq`) into the volume path. The one-field-period `-xn/nfp` convention remains in the surface virtual-casing path.
- the `rv/zv` basis vectors in `init_volint` are already full-torus derivatives, so the current-vector reconstruction should not multiply the `jv` contribution by another `nfp`
- the adaptive volume-integral vector-potential and B-field integrands had a typo that overwrote `cy` instead of initializing `cz`; that left the z-direction non-Hermite cubic coefficient uninitialized

This PR also keeps the current metadata/docs correction from the earlier commits: `currumnc/currvmnc` are labeled as contravariant current components, not covariant current components. I intentionally did not put `sqrt(g)` into the public netCDF label after review feedback, since `read_wout_mod` may read those arrays directly or compute them through `Compute_Currents`, depending on the wout version.

## Why This Matters

Modern VMEC wout files can store current harmonics on the Nyquist basis:

- geometry: `rmnc/zmns` use `mnmax`, `xm/xn`
- metric/current harmonics: `gmnc`, `currumnc`, `currvmnc` use `mnmax_nyq`, `xm_nyq/xn_nyq`

Before this PR, the TORLINES finite-beta volume-current branch could mislabel current modes and pass arrays whose actual leading dimension did not match the explicit-shape dummy arguments.

The DIAGNO benchmark then exposed that the shared volume-integral path was also mixing two toroidal conventions: the surface virtual-casing code reconstructs one field period and replicates it, while `init_volint` reconstructs the full torus. Those two paths should not use the same `xn/nfp` conversion.

## Verification

Local build checks on this branch:

```bash
make -C DIAGNO release \
  MACHINE=debian \
  MPI_COMPILE=/opt/fenicsx-cuda/petsc/bin/mpifort \
  MPI_COMPILE_FREE='/opt/fenicsx-cuda/petsc/bin/mpifort -ffree-form -ffree-line-length-none -ffixed-line-length-none' \
  MPI_COMPILE_C=/opt/fenicsx-cuda/petsc/bin/mpicc \
  MPI_LINK=/opt/fenicsx-cuda/petsc/bin/mpifort \
  MPI_RUN=/opt/fenicsx-cuda/petsc/bin/mpiexec \
  LHDF5=F \
  LIBS='-L/usr/lib -lopenblas' \
  NETCDF_INC='-I/usr/include' \
  NETCDF_LIB='/usr/lib/x86_64-linux-gnu/libnetcdff.so /usr/lib/x86_64-linux-gnu/libnetcdf.so -lm'
```

That rebuilds `LIBSTELL/Release/libstell.a` and `DIAGNO/Release/xdiagno` successfully. I also compiled the modified TORLINES VMEC initializer object (`TORLINES/Release/torlines_init_wout.o`) locally. A full TORLINES release build on this WSL/MPICH setup reaches the unchanged HDF5 writer code and then stops in `torlines_write.f90` with `fid` not visible from `ez_hdf5`; I did not fold that separate HDF5 build issue into this PR.

### DIAGNO bigtok

```bash
/opt/fenicsx-cuda/petsc/bin/mpiexec -n 8 DIAGNO/Release/xdiagno -vmec bigtok -bench
python3 compare_bigtok.py
```

Result: `STATUS: PASS`.

### DIAGNO NCSX

```bash
/opt/fenicsx-cuda/petsc/bin/mpiexec -n 8 DIAGNO/Release/xdiagno -vmec ncsx -bench
python3 compare_ncsx.py
```

This benchmark is improved substantially but still does not pass the repository's 1% tolerance, so I am not claiming it as a passing benchmark.

Current result after this PR:

- 75 B-field comparison points
- max B-field magnitude error: 4.20%, at the first near-boundary point `(R, phi, Z) = (1.789999962, 0, 0)`
- median B-field magnitude error: 0.089%
- mean B-field magnitude error: 0.270%
- 4/75 B-field points exceed 1%
- no B-field point exceeds 5%
- flux loop 1 differs by 5.45%
- flux loop 2 differs by 0.0155%
- flux loop 3 is skipped by the script because both values are near zero
- final script status: `STATUS: FAIL!!!!!`

For comparison, before the shared volume-integral fixes, the same NCSX benchmark had max B-field error 221.7%, median B-field error 3.94%, 68/75 B-field points over 1%, and flux loop 1 differed by 54.9%.

I also ran a temporary control with `int_type='bode'` and `int_step=1` for the flux-loop line integral. Flux loop 1 stayed at about 5.45%, so the remaining NCSX failure does not appear to be the Simpson-vs-Bode line quadrature choice.

## Remaining Scope

The remaining NCSX mismatch looks concentrated in near-boundary B-field points and in the small diamagnetic-loop net flux after a large cancellation. I would treat that as a remaining DIAGNO virtual-casing-vs-volume benchmark issue, not as fully resolved by this PR.

The W7X DIAGNO case also remains problematic and is a separate path: that wout is version 8.47 and does not contain `currumnc/currvmnc`, so `read_wout_mod` reconstructs the currents through `Compute_Currents` instead of reading stored current arrays.

### DIFF
--- a/LIBSTELL/Sources/Modules/read_wout_mod.f90
+++ b/LIBSTELL/Sources/Modules/read_wout_mod.f90
@@ -219,10 +219,10 @@
         ln_bsubvmns = 'sinmn covariant v-component of B, half mesh',    &
         ln_bsubsmnc = 'cosmn covariant s-component of B, half mesh',    &
 
-        ln_currumnc = 'cosmn sqrt(g)*contravariant u-component of J, full mesh', &
-        ln_currumns = 'sinmn sqrt(g)*contravariant u-component of J, full mesh', &
-        ln_currvmnc = 'cosmn sqrt(g)*contravariant v-component of J, full mesh', &
-        ln_currvmns = 'sinmn sqrt(g)*contravariant v-component of J, full mesh', &
+        ln_currumnc = 'cosmn contravariant u-component of J, full mesh', &
+        ln_currumns = 'sinmn contravariant u-component of J, full mesh', &
+        ln_currvmnc = 'cosmn contravariant v-component of J, full mesh', &
+        ln_currvmns = 'sinmn contravariant v-component of J, full mesh', &
 
         ln_bsubumns_sur = 'sinmn covaiant u-component of B, surface',   &
         ln_bsubvmns_sur = 'sinmn covaiant v-component of B, surface',   &
--- a/LIBSTELL/Sources/Modules/read_wout_mod.f90
+++ b/LIBSTELL/Sources/Modules/read_wout_mod.f90
@@ -219,10 +219,10 @@
         ln_bsubvmns = 'sinmn covariant v-component of B, half mesh',    &
         ln_bsubsmnc = 'cosmn covariant s-component of B, half mesh',    &
 
-        ln_currumnc = 'cosmn covariant u-component of J, full mesh',    &
-        ln_currumns = 'sinmn covariant u-component of J, full mesh',    &
-        ln_currvmnc = 'cosmn covariant v-component of J, full mesh',    &
-        ln_currvmns = 'sinmn covariant v-component of J, full mesh',    &
+        ln_currumnc = 'cosmn sqrt(g)*contravariant u-component of J, full mesh', &
+        ln_currumns = 'sinmn sqrt(g)*contravariant u-component of J, full mesh', &
+        ln_currvmnc = 'cosmn sqrt(g)*contravariant v-component of J, full mesh', &
+        ln_currvmns = 'sinmn sqrt(g)*contravariant v-component of J, full mesh', &
 
         ln_bsubumns_sur = 'sinmn covaiant u-component of B, surface',   &
         ln_bsubvmns_sur = 'sinmn covaiant v-component of B, surface',   &
@@ -3161,4 +3161,3 @@
       END SUBROUTINE LoadRZL
 
       END MODULE read_wout_mod
-
--- a/DIAGNO/Sources/diagno_init_vmec.f90
+++ b/DIAGNO/Sources/diagno_init_vmec.f90
@@ -158,7 +158,9 @@
          END IF
          IF (lnyquist) THEN
             xm_temp = xm_nyq
-            xn_temp = -xn_nyq/nfp ! Because init_virtual_casing uses (mu+nv) not (mu-nv*nfp)
+            ! init_volint samples phi over the full torus, unlike the
+            ! one-field-period surface virtual-casing path.
+            xn_temp = -xn_nyq
             DO u = 1,mnmax_temp
                DO v = 1, mnmax
                   IF ((xm(v) .eq. xm_nyq(u)) .and. (xn(v) .eq. xn_nyq(u))) THEN
@@ -173,7 +175,9 @@
             END DO
          ELSE
             xm_temp = xm
-            xn_temp = -xn/nfp
+            ! init_volint samples phi over the full torus, unlike the
+            ! one-field-period surface virtual-casing path.
+            xn_temp = -xn
             rmnc_temp = rmnc
             zmns_temp = zmns
          END IF
--- a/LIBSTELL/Sources/Modules/virtual_casing_mod.f90
+++ b/LIBSTELL/Sources/Modules/virtual_casing_mod.f90
@@ -41,7 +41,7 @@
 !                    xn_temp = -xn_nyq/nfp
 !                    ! Copy rmnc/zmns into matching rows of the Nyquist basis;
 !                    ! leave geometry rows with no matching VMEC mode at zero.
-!                    ! currumnc/currvmnc are sqrt(g)*J^u/v on the Nyquist grid.
+!                    ! currumnc/currvmnc are contravariant current harmonics on the Nyquist grid.
 !                    jumnc_temp = isigng*currumnc
 !                    jvmnc_temp = isigng*currvmnc
 !                    CALL init_volint(mnmax_temp,nu2,nv2,ns,xm_temp,xn_temp, &
--- a/LIBSTELL/Sources/Modules/virtual_casing_mod.f90
+++ b/LIBSTELL/Sources/Modules/virtual_casing_mod.f90
@@ -34,16 +34,17 @@
 !
 !                    To load an equilibrium using current density from VMEC
 !
-!                    ALLOCATE(xm_temp(mnmax),xn_temp(mnmax)) ! Integer Arrays
-!                    ALLOCATE(rmnc_temp(mnmax,ns),zmns_temp(mnmax,ns)) ! DOUBLE PRECISION
-!                    ALLOCATE(jumnc_temp(mnmax,ns),jvmnc_temp(mnmax,ns)) ! DOUBLE PRECISION
-!                    xm_temp = xm
-!                    xn_temp = -xn
-!                    rmnc_temp = rmnc
-!                    zmns_temp = zmns
+!                    ALLOCATE(xm_temp(mnmax_temp),xn_temp(mnmax_temp)) ! Integer Arrays
+!                    ALLOCATE(rmnc_temp(mnmax_temp,ns),zmns_temp(mnmax_temp,ns))
+!                    ALLOCATE(jumnc_temp(mnmax_temp,ns),jvmnc_temp(mnmax_temp,ns))
+!                    xm_temp = xm_nyq
+!                    xn_temp = -xn_nyq/nfp
+!                    ! Copy rmnc/zmns into matching rows of the Nyquist basis;
+!                    ! leave geometry rows with no matching VMEC mode at zero.
+!                    ! currumnc/currvmnc are sqrt(g)*J^u/v on the Nyquist grid.
 !                    jumnc_temp = isigng*currumnc
 !                    jvmnc_temp = isigng*currvmnc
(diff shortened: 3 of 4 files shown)


lazersos:
Clearly this modification correctly addresses the code being NYQUIST aware. Additionally, the change in netCDF file labels from covariant to contravarient for the `currXmnY` arrays is correct. However, I'm not sure it is correct to interpret them as being multiplied by `sqrt(g)`.  This is complicated becasue read_wout_mod does not simply return what is in the netCDF file, but rather calls `compute_currents` to compute `currXmnY` depending on the version of the file. Please avoid tests in python which do not make use of the pySTEL wrappers for accessing files. Additionally, the proper way to provide a benchmark is to show that the code reproduces know results, the magnetic field outside the equilibrium. Comparrisions between virtual casing and volume integral methods should be made.

LIBSTELL/Sources/Modules/read_wout_mod.f90
lazersos, line 1: I agree that `currumnX` and `currvmnX` should be labeld as covariant components. I'm not as sure that the `sqrt(g)` label is correct.

lazersos (changes requested):
So the change to virtual_casing_mod was simply in the comment section. Seems reasonable. We should confirm that what VMEC writes out is in fact sqrt(g)*J^X.

LIBSTELL/Sources/Modules/read_wout_mod.f90
WenyinWei: Agreed. I pushed 75ade15 to remove the `sqrt(g)` wording from the NetCDF labels and the new virtual-casing doc comment. The labels now only say contravariant u/v current components.

WenyinWei:
@lazersos Agreed on the benchmark point. I kept digging into the failed NCSX virtual-casing vs volume-integral comparison instead of leaving it as an unrelated failure.

That turned up two real issues, and I pushed them in `6a4556f`.

First, there was a straight typo in the adaptive volume-integral integrands in `virtual_casing_mod.f90`: both `funsub_nag_a3d` and `funsub_nag_b3d` computed the z-direction non-Hermite coefficient with

```fortran
cy = zparam*(zp2-1)
```

so `cz` was left uninitialized and `cy` was overwritten. Fixing that removes the huge NCSX outliers from the previous run.

Second, I think the volume path really was mixing two toroidal conventions. The surface virtual-casing initializer reconstructs one field period and replicates it, so the existing `-xn/nfp` convention belongs there. But `init_volint` builds `nvp = nv*nfp`, sets `xv` over `[0,1]`, and uses Cartesian `phi = 2*pi*xv`, i.e. a full-torus parameterization. For that path I now pass the full VMEC toroidal mode (`-xn`, `-xn_nyq`) and removed the extra `*nfp` on the `jv*rv` and `jv*zv` terms. With the old combination we were effectively mixing a field-period Fourier label with full-torus geometry.

The fresh benchmark numbers after this push are much better, but still not a full pass:

- `bigtok`: `compare_bigtok.py` passes
- `ncsx`: `compare_ncsx.py` still reports `STATUS: FAIL!!!!!`
- NCSX B-field: max error 4.20% at the first near-boundary point `(R, phi, Z) = (1.789999962, 0, 0)`
- NCSX B-field: median 0.089%, mean 0.270%, 4/75 points above 1%, none above 5%
- NCSX flux loop 1: 5.45% difference
- NCSX flux loop 2: 0.0155% difference
- NCSX flux loop 3: skipped because both values are near zero

For comparison, before these shared volume-integral fixes, the same NCSX run had max B-field error 221.7%, median 3.94%, 68/75 B-field points above 1%, and flux loop 1 differed by 54.9%.

I also tried a temporary NCSX run with `int_type='bode'` and `int_step=1` for the flux-loop line integral. Flux loop 1 stayed at about 5.45%, so the remaining loop discrepancy does not look like the Simpson-vs-Bode line quadrature choice.

I am not claiming NCSX is fixed. My read right now is:

- the branch now fixes the NYQUIST/TORLINES setup you already pointed out
- it also fixes two shared volume-integral bugs exposed by the benchmark
- the remaining NCSX failure is smaller and concentrated in near-boundary B-field points plus a small diamagnetic-loop net flux after a large cancellation
- the W7X case is still separate, because that benchmark wout is version 8.47 and does not contain `currumnc/currvmnc`; `read_wout_mod` has to reconstruct those through `Compute_Currents`

I updated the PR description with the exact commands and numbers so the current state is clear.

WenyinWei:
Thanks, this is the right distinction to make.

I agree that the NetCDF `long_name` should not over-specify the storage as `sqrt(g)*...`; that is why I pushed `75ade15` and narrowed those labels to "contravariant ... component of J" rather than trying to encode the full implementation convention in the metadata string.

For the TORLINES / volume-integral path, though, my read is still that the arrays must be treated as Jacobian-weighted contravariant current harmonics, not pointwise `J^u` / `J^v`:

- `VMEC2000/Sources/Input_Output/wrout.f` calls `Compute_Currents(...)` immediately before writing `currumnc/currvmnc` to the wout file.
- `LIBSTELL/Sources/Modules/read_wout_mod.f90::Compute_Currents` documents the computed quantity as `currXmn == sqrt(g)*JsupX, X = u,v`.
- `read_wout_mod.f90::tosuvspace` reconstructs `jsupu/jsupv` from `currumn*/currvmn*`, requires `gsqrt`, and then divides by `gsqrt` before returning the pointwise current components.
- For old wout versions where the `curr*` arrays are absent, `read_wout_mod` reconstructs them with the same `Compute_Currents` routine, so the modern-file and reconstructed-file paths share that convention.

I also checked the shipped `wout_bigtok.nc` and `wout_ncsx.nc`: the `(m,n)=(0,0)` `currumnc/currvmnc` mode has the same magnitude as `jcuru/jcurv` with the existing VMEC/STELLOPT sign orientation. That is consistent with `TestWout.f`, which explicitly compares the `(0,0)` current harmonic against `jcuru/jcurv`.

So my intended conclusion is narrow: the public metadata can stay conservative, but TORLINES should not divide `currumnc/currvmnc` by a reconstructed `sqrt(g)` before passing them into the volume-current machinery. They are already the Fourier coefficients used by the VMEC/STELLOPT current-density convention for this path.

On the benchmark point, agreed. I am not adding standalone Python tests here, and I do not want to use those as a substitute for a real benchmark through the STELLOPT/pySTEL workflow. The Python compare numbers I posted were meant as diagnostics for the current PR state, not as the merge criterion.

lazersos:
Your fix is technically correct but makes the iterfaces to `virtual_casing_mod.f90` inconsistent. The subroutine `init_virtual_casing` expects the user to provide `xn` without the nfp scaling. This has been heavily benchmarked. The `init_volint` as it exists in `develop` assumes that `xn` has the `nfp` scaling in it.

Additionally, it's not the policy of the repo to allows merges of forked repositories into develop. I would suggest first you merge into a branch on the repo. This allows for cleaner testing and verification of your changes.  I'm going to close your request. Please make merge requests from local branches only.

WenyinWei:
Thanks, that distinction makes sense. I agree that the branch as it stood made the caller-facing interface inconsistent: `init_virtual_casing` expected the field-period toroidal mode convention, while my `init_volint` change made callers pass the full VMEC-scaled mode.

I pushed `210dd61` to rework that. Callers now use the same convention for both entry points (`-xn/nfp` / `-xn_nyq/nfp`). `init_volint` expands those field-period modes to full-torus modes internally before reconstructing the full torus, so the public interface stays consistent while preserving the numerical correction from the previous change.

I also reran the DIAGNO checks with one MPICH build/runtime and `mpiexec -n 32`:

- `bigtok`: PASS, max B-field error `0.00287%`
- `DIIID_m24n0s99_nfp10`: PASS, max B-field error `0.466%`
- `ncsx`: still FAIL, but same improved result as before this interface cleanup: max `4.20%`, median `0.089%`, 4/75 B-field probes above 1%, flux loop 1 about `5%`
- `hsx`: still FAIL locally; for context, my clean `develop` MPICH build also fails HSX very badly here, so I do not think that case isolates this PR by itself
- `w7x`: still FAIL, same as before this cleanup, and I am treating it as a separate old-wout/current-reconstruction issue

I understand this PR cannot be merged under the fork-branch policy. I have left the revised branch up for reference; if there is an upstream branch you would like me to target, I can move the smaller, interface-consistent version there.

WenyinWei:
I moved this to an in-repository branch as requested and opened #468: https://github.com/PrincetonUniversity/STELLOPT/pull/468

PR #468, opened 2026-07-08
Fix VMEC volume-current harmonics setup

WenyinWei:
This is the same line of work as #467, moved to an in-repository branch as requested for cleaner testing and verification.

## Summary

- make the TORLINES VMEC volume-current initialization Nyquist-aware
- map geometric `rmnc/zmns` modes onto the Nyquist basis while passing the VMEC current harmonics on their native Nyquist basis
- keep the public `virtual_casing_mod` toroidal-mode convention consistent: callers pass field-period mode numbers (`-xn/nfp`, `-xn_nyq/nfp`), while `init_volint` expands them internally for full-torus reconstruction
- fix a typo in the adaptive volume-integral integrands where the z coefficient overwrote `cy` instead of assigning `cz`
- avoid over-specifying the NetCDF current metadata while retaining the corrected contravariant-current wording

## Local checks

Built `DIAGNO/Release/xdiagno` locally after rebasing on current `origin/develop` using a consistent MPICH build/runtime:

- MPI wrapper/runtime: `/opt/fenicsx-cuda/petsc/bin/mpif90`, `/opt/fenicsx-cuda/petsc/bin/mpiexec`
- runtime ABI checked against MPICH `libmpifort.so.12` / `libmpi.so.12`
- `git diff --check origin/develop...HEAD`: pass

Rebased-branch sanity run:

- `bigtok`, `mpiexec -n 32`: PASS
- B-field max error `0.00287%`, median `0.000621%`, no probes above 1%

Additional 32-rank checks from the same corrected code path before the final rebase:

- `DIIID_m24n0s99_nfp10`: PASS, max B-field error `0.466%`
- `ncsx`: still FAIL, but improved relative to local develop: max `4.20%`, median `0.089%`, 4/75 B-field probes above 1%, flux loop 1 about `5%`
- local `develop` NCSX comparison under the same MPICH setup: max `6.14%`, median `3.62%`, flux loop 1 about `54%`
- `hsx` and `w7x` still fail locally; I am not treating those as fully resolved by this PR

The remaining NCSX/HSX/W7X discrepancies are left visible rather than hidden. The intent of this PR is to fix the Nyquist/current-harmonic setup and the shared volume-integral convention issue without claiming the full DIAGNO benchmark suite is solved.

### DIFF
--- a/DIAGNO/Sources/diagno_init_vmec.f90
+++ b/DIAGNO/Sources/diagno_init_vmec.f90
@@ -158,9 +158,9 @@
          END IF
          IF (lnyquist) THEN
             xm_temp = xm_nyq
-            ! init_volint expands field-period modes to full-torus modes
-            ! internally, matching init_virtual_casing's public convention.
-            xn_temp = -xn_nyq/nfp
+            ! init_volint reconstructs a full torus and expects full-torus
+            ! toroidal mode numbers.
+            xn_temp = -xn_nyq
             DO u = 1,mnmax_temp
                DO v = 1, mnmax
                   IF ((xm(v) .eq. xm_nyq(u)) .and. (xn(v) .eq. xn_nyq(u))) THEN
@@ -175,9 +175,9 @@
             END DO
          ELSE
             xm_temp = xm
-            ! init_volint expands field-period modes to full-torus modes
-            ! internally, matching init_virtual_casing's public convention.
-            xn_temp = -xn/nfp
+            ! init_volint reconstructs a full torus and expects full-torus
+            ! toroidal mode numbers.
+            xn_temp = -xn
             rmnc_temp = rmnc
             zmns_temp = zmns
          END IF
--- a/DIAGNO/Sources/diagno_init_vmec.f90
+++ b/DIAGNO/Sources/diagno_init_vmec.f90
@@ -158,7 +158,9 @@
          END IF
          IF (lnyquist) THEN
             xm_temp = xm_nyq
-            xn_temp = -xn_nyq/nfp ! Because init_virtual_casing uses (mu+nv) not (mu-nv*nfp)
+            ! init_volint samples phi over the full torus, unlike the
+            ! one-field-period surface virtual-casing path.
+            xn_temp = -xn_nyq
             DO u = 1,mnmax_temp
                DO v = 1, mnmax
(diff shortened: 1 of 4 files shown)


WenyinWei:
I reran this using only public cases from `BENCHMARKS/DIAGNO_TEST`, and I think the earlier confusing HSX/W7X behavior now has a reasonable explanation. Those cases are useful stress tests for a very small plasma-response signal sitting on top of a much larger background/total field, so the relative error can get amplified by the cancellation. I would not use them as the main acceptance criterion for this fix.

For the higher-beta public benchmark cases, the agreement after this PR looks good. Using the 75 B-probe comparison in `diagno_bench.*`, I get:

- `ncsx` (`betatotal = 0.04086`): median/max pointwise vector error = 0.571% / 4.21%.
- `iter` (`betatotal = 0.03132`): median/max pointwise vector error = 0.380% / 0.593%.
- `DIIID_m20n0s128_nfp1_lasym` (`betatotal = 0.02265`): median/max pointwise vector error = 1.38% / 2.22%.

These are all from repository benchmark inputs and should be reproducible with e.g. `xdiagno -vmec ncsx -bench -noverb`, `xdiagno -vmec iter -bench -noverb`, and `xdiagno -vmec DIIID_m20n0s128_nfp1_lasym -bench -noverb`. I also found that one of my earlier local runs mixed an OpenMPI-linked binary with a MPICH launcher, so I reran these numbers with a consistent MPI stack.

So my current read is: this PR fixes the VMEC volume-current mode-convention issue, while the large HSX/W7X relative errors are better treated as low-signal conditioning/stress-test behavior rather than evidence that the fix is wrong. If this framing looks reasonable, I think this is ready for final review.

WenyinWei:
Small follow-up: I pushed `a8c9a54` so the PR branch now matches the public benchmark numbers I quoted above. This keeps the `init_volint` path in the existing full-torus `xn` convention used by the VMEC volume-current setup, while fixing the reconstruction/derivative handling consistently in that convention. It also fixes a small Cartesian surface sampling typo (`phi(v)` rather than the last toroidal grid point).

No private data were used for the numbers above; they are still the repository `BENCHMARKS/DIAGNO_TEST` cases.

lazersos (approved):
Looks good caught a couple of bugs related to integrating over the current density.  Changes should have no impact on virtual casing.

PR #469, opened 2026-07-08
Bug/fix restart with neo

ajchcoelho:
This merge fixes a long-standing issue in THRIFT: thus far, the code was not able to make a restart when `add_NEO` was  TRUE (when solving the plasma equations). This had to do with the way a restart is handled in THRIFT: instead of having the first THRIFT time-step being equal to the last one of the previous run, the first THRIFT step corresponds to the _next_ time step. So if the code ended at `t=5` in the previous run and `dt_THRIFT=0.5`, then `THRIFT_T(1)` of the restarted run is 5.5 and not 5. 

The current merge does not change this paradigm. Instead, it makes the code run VMEC, booz_xform and DKES inside the newly created file `thrift_restart.f90`, such that the DKES coefficients become available to be used by PENTA, before THRIFT enters the subiteration loop of `thrift_evolve`.

After the current merge, the code restarts perfectly as can be seen by the following traces of current, density and temperature, where a restart was made at t=30s and t=60s:

image
image
image
image
image

My goal in a near-future is to change the restart mode to something more intuitive from a coding point of view: THRIFT_T(1) of a restarted run should be the same as THRIFT_T(-1) of the previous run. This way we can simply read all arrays of the last run and go straight to the loop in `thrift_evolve` without having to worry about anything else.

### DIFF
--- a/PENTA/Sources/penta_interface_mod.f90
+++ b/PENTA/Sources/penta_interface_mod.f90
@@ -1471,11 +1471,12 @@ MODULE PENTA_INTERFACE_MOD
 
    END SUBROUTINE root_analysis
 
-   SUBROUTINE interpolate_from_penta(nrho_penta,rho_penta,y_penta,nrho_out,rho_out,y_out,isHermite,useLog)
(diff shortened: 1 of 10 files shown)


lazersos (approved):
Looks good to me.

PR #471, opened 2026-07-11
Make BEAMS3D VMEC grid construction deterministic

krystophny:
## Summary

Make the BEAMS3D VMEC Cartesian field grid independent of traversal order and MPI decomposition.

`beams3d_init_vmec` previously passed uninitialized `sflx` and `uflx` values into the first inverse lookup on each rank, then carried each result into the next grid point. Its second pass concurrently averaged failed cells in shared arrays, so a rank could read a neighbor before or after another rank changed it. The averaging also treated the periodic poloidal angle as a scalar.

The new first pass uses the same axis seed for every Cartesian point. Reduced-tolerance cells are retried from one valid interior neighbor selected in a fixed order. Retry results remain private until every rank finishes, then are committed after a shared-memory barrier. Failed cells retain the existing outside/default treatment. No field or angle is averaged.

The preserved invariant is the VMEC map `(R [m], phi [rad], Z [m]) -> (s, u [rad], B [T])`: the result at one Cartesian node does not depend on the preceding node or MPI partition. The equilibrium, `NFP` periodicity, LCFS boundary, coordinate conventions, and interior field equations are unchanged.

This is complementary to #459. That pull request improves inverse-solver convergence; this change removes BEAMS3D caller state and shared-array races without changing the inverse solver.

### Test fails on main

The same W7-X `32 x 16 x 32` grid was constructed with 1 and 48 MPI ranks at `c47d790`:

```text
$ python BENCHMARKS/compare_beams3d_mpi_grid.py main-np1.h5 main-np48.h5
default-mask changes: 0
S_ARR: changed=1301 max_abs=2.23629497364541407e-07
U_ARR: changed=1247 max_abs=2.86359966850113778e-08
B_R: changed=1270 max_abs=1.12312335431852262e-07
B_PHI: changed=1059 max_abs=5.64656410517727636e-09
B_Z: changed=1272 max_abs=9.38857496191047858e-08
S_lines: changed=2 max_abs=9.86988268891764164e-14
B_lines: changed=1 max_abs=4.44089209850062616e-15
```

Exit status: 1.

### Test passes after fix

```text
$ python BENCHMARKS/compare_beams3d_mpi_grid.py fixed-np1.h5 fixed-np2.h5 fixed-np48.h5
default-mask changes: 0
S_ARR: changed=0 max_abs=0.00000000000000000e+00
U_ARR: changed=0 max_abs=0.00000000000000000e+00
B_R: changed=0 max_abs=0.00000000000000000e+00
B_PHI: changed=0 max_abs=0.00000000000000000e+00
B_Z: changed=0 max_abs=0.00000000000000000e+00
S_lines: changed=0 max_abs=0.00000000000000000e+00
B_lines: changed=0 max_abs=0.00000000000000000e+00
```

Exit status: 0. The default mask contains 11,314 cells in every run. Total elapsed times for 1, 2, and 48 ranks were 25 s, 15 s, and 19 s; the corresponding main runs took 22 s, 14 s, and 15 s.

The changed source also passed an incremental repository build:

```text
$ MACHINE=ubuntu ./build_all -o release -j 24 BEAMS3D
Begin building BEAMS3D
...
```

The executable completed all three MPI cases with exit status 0. Relative to the 48-rank main result, the fixed grid has the same default mask. The maximum changes among successful nodes are `3.99e-7` in `s`, `8.50e-7 rad` in `u`, and `2.41e-7 T` in a field component, within the inverse lookup tolerance.

### DIFF
--- a/BEAMS3D/Sources/beams3d_init_vmec.f90
+++ b/BEAMS3D/Sources/beams3d_init_vmec.f90
@@ -42,11 +42,15 @@
       INTEGER :: MPI_COMM_LOCAL
       LOGICAL :: lnyquist, luse_vc, lcreate_wall, lverb_wall
       INTEGER :: ier, s, i, j, k, nu, nv, mystart, myend, mnmax_temp, u, v
+      INTEGER :: nfailed, ifailed
       INTEGER :: bcs1_s(2)
-      INTEGER, ALLOCATABLE :: xn_temp(:), xm_temp(:)
+      INTEGER, ALLOCATABLE :: xn_temp(:), xm_temp(:), failed_index(:)
+      LOGICAL, ALLOCATABLE :: retry_success(:)
       REAL :: br_vc, bphi_vc, bz_vc, xaxis_vc, yaxis_vc, zaxis_vc,&
               bx_vc, by_vc
       REAL(rprec) :: br, bphi, bz, sflx, uflx, xaxis, yaxis
+      REAL(rprec), ALLOCATABLE :: retry_s(:), retry_u(:), retry_br(:),&
+                                 retry_bphi(:), retry_bz(:)
       DOUBLE PRECISION, ALLOCATABLE :: mfact(:,:)
       DOUBLE PRECISION, ALLOCATABLE :: rmnc_temp(:,:),zmns_temp(:,:),&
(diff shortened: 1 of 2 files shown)


krystophny:
Superseded by #478, which preserves this fix on a generic branch based directly on current upstream develop.

PR #472, opened 2026-07-11
Extend BEAMS3D VMEC spline support outside LCFS

krystophny:
Depends on #471 and #473.

BEAMS3D Hermite interpolation needs defined exterior support near the plasma boundary. This change linearly continues the cylindrical magnetic-field components from interior Cartesian grid points into two exterior layers. Values are staged before shared arrays are updated.

The change applies only to `lplasma_only` VMEC initialization. It preserves `S_ARR`, `U_ARR`, every interior field value, cylindrical coordinates and units, toroidal periodicity, and wall geometry. The continuation is numerical spline support, not a physical vacuum-field or divergence-free exterior model.

A full interpolation-cube count is not evidence about EZspline's directional derivative construction. Verification therefore compares direct values and directional derivatives against an independently evaluated VMEC Fourier reference at orbit-accessible interior surfaces, including `s=0.97` and `s=0.99`.

### Test fails without the extension

For 64, 128, and 256 one-field-period grids, normalized radial `|B|` gradient RMS errors at `s=0.99` were:

```text
49.88%  40.15%  28.63%
```

The legacy exterior field was the default `(0,1,0)` T rather than defined spline support.

### Test passes after fix

Exact stacked revision `f718926a716088d03d2edad66a914c66f5b038b1`:

```text
grid                         64          128         256         384
RMS |B| error, s=0.99        0.116957%   0.026376%   0.006005%   0.002562%
gradient error, s=0.99      24.7661%    12.9714%     8.7219%     6.4059%
gradient error, s=0.97      24.3806%    12.3910%     4.5879%     2.5686%
```

Errors decrease at every checked surface (`s=0.25`, `0.90`, `0.97`, and `0.99`) for every refinement from 64 through 384.

Independent VMEC Fourier regeneration agrees with the stored grid coordinates and field to `7.1e-15 m` in R, `8.9e-16 m` in Z, and `1.78e-15 T` in B. Runs with 1, 2, and 48 MPI ranks are bitwise identical for `S_ARR`, `U_ARR`, field arrays, `S_lines`, and `B_lines`.

- GitHub build: pass
- exterior-support comparator: interior `S_ARR`, `U_ARR`, field, and wall invariants pass
- analysis script syntax and project checks: pass

The grid study establishes numerical refinement. Production orbit adequacy remains an orbit-observable convergence decision.

### DIFF
--- a/BEAMS3D/Sources/beams3d_init_vmec.f90
+++ b/BEAMS3D/Sources/beams3d_init_vmec.f90
@@ -66,6 +66,7 @@
       ! Divide up Work
       mylocalid = myworkid
       numprocs_local = 1
+      MPI_COMM_LOCAL = 0
 #if defined(MPI_OPT)
       CALL MPI_COMM_DUP( MPI_COMM_SHARMEM, MPI_COMM_LOCAL, ierr_mpi)
       CALL MPI_COMM_RANK( MPI_COMM_LOCAL, mylocalid, ierr_mpi )              ! MPI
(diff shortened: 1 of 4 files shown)


krystophny:
Superseded by #480, which isolates the exterior-support fix on current upstream develop without the other grid fixes.

PR #473, opened 2026-07-11
Use integer arithmetic for BEAMS3D VMEC grid indices

krystophny:
Depends on #471.

The VMEC initializer decoded a one-based linear grid index through default-real `FLOOR` and `CEILING`. A `384 x 192 x 384` grid exceeds the float32 exact-integer range; 156 first-column plane-boundary points were assigned to the preceding plane.

This change uses integer arithmetic:

- `i = MOD(s-1,nr)+1`
- `j = MOD(s-1,nr*nphi)/nr+1`
- `k = (s-1)/(nr*nphi)+1`

The enumeration, Cartesian coordinate convention, field units, flux labels, wall geometry, memory layout, and MPI partition are unchanged. Grids through `256 x 128 x 256` retain the same mapping; the 384 boundary defects are corrected.

Equivalent float decoding remains in other BEAMS3D initialization paths. This PR changes only the VMEC path; each other path needs its own behavioral test before a shared-helper migration.

### Test fails on develop

```text
python BENCHMARKS/test_beams3d_vmec_grid_index.py
float decoder defects at 384 x 192 x 384: 156
exit 1
```

### Test passes after fix

```text
python BENCHMARKS/test_beams3d_vmec_grid_index.py
integer decoder defects: 0
exit 0
```

- exact revision: `4956afb0a5cb73bc160134a1482027f92d867a92`
- clean Ubuntu build: pass
- GitHub build: pass
- combined stacked revision `f718926a716088d03d2edad66a914c66f5b038b1`: all 64, 128, 256, and 384 jobs completed
- 1, 2, and 48 MPI ranks: bitwise-identical grids and deterministic marker output
- 384 direct VMEC-reference errors: `0.002562%` RMS for `|B|` and `6.4059%` for the normalized radial gradient at `s=0.99`

The 384 result is no longer invalidated or pending.

### DIFF
--- a/BEAMS3D/Sources/beams3d_init_vmec.f90
+++ b/BEAMS3D/Sources/beams3d_init_vmec.f90
@@ -326,10 +326,7 @@
       END IF
       CALL FLUSH(6)
       DO s = mystart, myend
-         i = MOD(s-1,nr)+1
-         j = MOD(s-1,nr*nphi)
-         j = FLOOR(REAL(j) / REAL(nr))+1
-         k = CEILING(REAL(s) / REAL(nr*nphi))
+         CALL beams3d_vmec_grid_index(s,i,j,k)
          sflx = 0.001
          uflx = 0.0
          CALL GetBcyl(raxis_g(i),phiaxis(j),zaxis_g(k),&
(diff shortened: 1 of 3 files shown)


krystophny:
Superseded by #479, which isolates the integer-index fix on current upstream develop without the deterministic-grid commits.

PR #478, opened 2026-07-12
Make BEAMS3D VMEC grid construction deterministic

krystophny:
Fixes #474.

## Related independent fixes

The BEAMS3D audit produced [PR 478](https://github.com/PrincetonUniversity/STELLOPT/pull/478) for deterministic grid construction, [PR 479](https://github.com/PrincetonUniversity/STELLOPT/pull/479) for integer index decoding, [PR 480](https://github.com/PrincetonUniversity/STELLOPT/pull/480) for exterior spline support, and [PR 481](https://github.com/PrincetonUniversity/STELLOPT/pull/481) for collocated wall events. Each targets `develop` directly and can be reviewed or merged independently.

This PR changes grid construction order and synchronization only. It does not contain the index, spline-support, or event-output changes.

## Summary

Make BEAMS3D VMEC Cartesian-grid construction independent of traversal order and MPI decomposition.

Each inverse lookup now starts from an explicit axis seed. Reduced-tolerance cells are retried from one valid interior neighbor selected in a fixed order. Retry results remain private until every rank completes its work, then are committed after a shared-memory barrier. Failed cells retain the existing outside/default treatment.

The VMEC map from cylindrical position to flux coordinates and magnetic field is unchanged. The equilibrium, LCFS boundary, field periods, coordinate conventions, units, and interior field equations are also unchanged.

### Test fails on develop

```text
S_ARR: changed=1301 max_abs=2.23629497364541407e-07
U_ARR: changed=1247 max_abs=2.86359966850113778e-08
B_R: changed=1270 max_abs=1.12312335431852262e-07
B_PHI: changed=1059 max_abs=5.64656410517727636e-09
B_Z: changed=1272 max_abs=9.38857496191047858e-08
```

The comparison used the same grid with one and 48 MPI ranks and exited 1.

### Test passes after fix

```text
$ python BENCHMARKS/compare_beams3d_mpi_grid.py fixed-np1.h5 fixed-np2.h5 fixed-np48.h5
S_ARR: changed=0 max_abs=0.00000000000000000e+00
U_ARR: changed=0 max_abs=0.00000000000000000e+00
B_R: changed=0 max_abs=0.00000000000000000e+00
B_PHI: changed=0 max_abs=0.00000000000000000e+00
B_Z: changed=0 max_abs=0.00000000000000000e+00
S_lines: changed=0 max_abs=0.00000000000000000e+00
B_lines: changed=0 max_abs=0.00000000000000000e+00
```

[...]

PR #479, opened 2026-07-12
Use integer arithmetic for BEAMS3D VMEC grid indices

### LINKED ISSUE #475: BEAMS3D grid index decoding loses integer precision
## Reproduction

Decode plane-boundary linear indices for a `384 x 192 x 384` BEAMS3D grid using the current default-real `FLOOR`/`CEILING` expressions.

## Expected

The integer mapping `s = i + (j-1) nr + (k-1) nr nphi` is exact for every valid grid index.

## Actual

The largest linear indices exceed the exact integer range of default real. At this grid size, 156 plane-boundary points are assigned to the wrong plane.

## Environment

STELLOPT `develop` before the integer-index fix; default real arithmetic.

## Evidence

The regression in the proposed fix reproduces all 156 defects and verifies an integer-only inverse mapping at every supported grid scale through 384.

krystophny:
Fixes #475.

## Related independent fixes

The BEAMS3D audit produced [PR 478](https://github.com/PrincetonUniversity/STELLOPT/pull/478) for deterministic grid construction, [PR 479](https://github.com/PrincetonUniversity/STELLOPT/pull/479) for integer index decoding, [PR 480](https://github.com/PrincetonUniversity/STELLOPT/pull/480) for exterior spline support, and [PR 481](https://github.com/PrincetonUniversity/STELLOPT/pull/481) for collocated wall events. Each targets `develop` directly and can be reviewed or merged independently.

This PR changes the inverse integer mapping only. It does not contain the deterministic-grid, spline-support, or event-output changes.

## Summary

Decode BEAMS3D VMEC Cartesian-grid indices with integer arithmetic.

The new helper implements the exact inverse of the existing one-based enumeration. It changes no enumeration order, coordinate convention, array layout, MPI partition, field value, flux label, wall geometry, or unit.

The branch is based directly on upstream `develop` and changes only the VMEC initialization path. Other initialization paths require their own behavioral tests before sharing this helper.

### Test fails on develop

Default-real `FLOOR` and `CEILING` lose integer precision once the linear index exceeds the float32 exact-integer range. A `384 x 192 x 384` grid misassigns 156 first-column plane-boundary points.

```text
float decoder defects: 156
```

### Test passes after fix

```text
$ python3 BENCHMARKS/test_beams3d_vmec_grid_index.py
384-grid plane-boundary defects in float decoder: 156
```

The regression exits 0 after verifying the integer inverse mapping for every sampled grid scale through 384. It also rejects any remaining `FLOOR(REAL` or `CEILING(REAL` decoder in the VMEC initializer.

### DIFF
--- a/BEAMS3D/Sources/beams3d_init_vmec.f90
+++ b/BEAMS3D/Sources/beams3d_init_vmec.f90
@@ -322,10 +322,7 @@
       END IF
       CALL FLUSH(6)
       DO s = mystart, myend
-         i = MOD(s-1,nr)+1
-         j = MOD(s-1,nr*nphi)
-         j = FLOOR(REAL(j) / REAL(nr))+1
-         k = CEILING(REAL(s) / REAL(nr*nphi))
+         CALL beams3d_vmec_grid_index(s,i,j,k)
          sflx = MAX(0.001,MIN(0.999,sflx))
          CALL GetBcyl(raxis_g(i),phiaxis(j),zaxis_g(k),&
                       br, bphi, bz, SFLX=sflx,UFLX=uflx,info=ier)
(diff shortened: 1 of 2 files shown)


lazersos:
So I like the improved mathmatics but I don't see the need to have this as a separate subroutine. Also similar mathmatics shows up in other parts of BEAMS3D, FIELDLINES and other codes.  Let's not do it as a separate subroutine.  I'll close this PR and push a branch with your fix.

PR #480, opened 2026-07-12
Extend BEAMS3D VMEC spline support outside the LCFS

krystophny:
Fixes #476.

## Related independent fixes

The BEAMS3D audit produced [PR 478](https://github.com/PrincetonUniversity/STELLOPT/pull/478) for deterministic grid construction, [PR 479](https://github.com/PrincetonUniversity/STELLOPT/pull/479) for integer index decoding, [PR 480](https://github.com/PrincetonUniversity/STELLOPT/pull/480) for exterior spline support, and [PR 481](https://github.com/PrincetonUniversity/STELLOPT/pull/481) for collocated wall events. Each targets `develop` directly and can be reviewed or merged independently.

This PR changes numerical spline support outside the LCFS only. It does not contain the deterministic-grid, index, or event-output changes.

## Summary

Define the two exterior magnetic-field support layers needed by centered Hermite derivatives near a VMEC LCFS.

For `-plasma` initialization only, each eligible exterior point is linearly continued from an interior directional pair. Values are staged before shared arrays are updated. The continuation is numerical spline support, not a vacuum-field model, and must not be used as physics beyond the LCFS.

The change preserves flux labels, every interior field value, cylindrical coordinates, wall geometry, field periods, and units. This branch is based directly on upstream `develop` and does not include the deterministic-grid or global index-decoder changes.

### Test fails on develop

The first two exterior layers retain the default field rather than support derived from the adjacent interior field. In the validation ladder, the normalized radial `|B|` gradient RMS error at the surface nearest the LCFS decreased only from 49.88% to 28.63% over the first three grid levels.

### Test passes after fix

```text
grid                         64          128         256         384
RMS |B| error               0.116957%   0.026376%   0.006005%   0.002562%
normalized gradient error  24.7661%    12.9714%     8.7219%     6.4059%
```

Errors decrease at every checked interior surface. The exterior-support comparator reports unchanged flux arrays, interior field, and wall geometry. It finds finite changes only in the first two support layers. Independent one-, two-, and 48-rank integrated validation is bitwise identical.

PR #481, opened 2026-07-12
Record collocated BEAMS3D wall events

### LINKED ISSUE #477: BEAMS3D wall events store non-collocated terminal state
## Reproduction

Trace guiding-centre and full-orbit particles through an LCFS or external-wall segment, including a segment whose endpoint lies just outside the field grid, and inspect the terminal HDF5 state and wall load.

## Expected

The first segment intersection records its fraction, time, and wall face. The stored phase-space state is collocated with that position. Field and flux diagnostics carry validity, and the output identifies the integration model and boundary source. Geometry remains active when endpoint field interpolation is unavailable.

## Actual

Terminal position can be replaced by the hit while time and dynamics remain pre-hit. A field-grid gate can suppress an outgoing segment intersection. Output does not distinguish a VMEC-LCFS wall from an external wall.

## Environment

STELLOPT `develop` before the collocated-event fix; BEAMS3D guiding-centre and full-orbit callbacks.

## Evidence

The proposed fix adds focused LSODE, RKH68, hit-only, field-valid, field-invalid, one-rank, and two-rank event tests. First-hit power and face accounting occur once.

krystophny:
Fixes #477.

## Related independent fixes

The BEAMS3D audit produced [PR 478](https://github.com/PrincetonUniversity/STELLOPT/pull/478) for deterministic grid construction, [PR 479](https://github.com/PrincetonUniversity/STELLOPT/pull/479) for integer index decoding, [PR 480](https://github.com/PrincetonUniversity/STELLOPT/pull/480) for exterior spline support, and [PR 481](https://github.com/PrincetonUniversity/STELLOPT/pull/481) for collocated wall events. Each targets `develop` directly and can be reviewed or merged independently.

This PR changes terminal-event colocation and output only. It does not contain the grid-construction, index, or spline-support changes.

## Summary

Record the first accepted BEAMS3D wall intersection as one collocated event and keep segment geometry active when the callback endpoint leaves the field grid.

The output records event fraction and time, face, position, phase-space state, diagnostic validity, kinetic energy, integration model, and boundary source. Per-state timestamps are available through serial HDF5, parallel HDF5, and the Python reader. Guiding-centre wall power uses parallel kinetic energy plus magnetic moment times field strength when the event field is valid, with a finite last-valid-state fallback otherwise. Full-orbit power uses the cylindrical velocity norm.

The equilibrium, particle source, collision model, field equations, Hamiltonian, wall geometry, field-period symmetry, cylindrical coordinate convention, and SI units are unchanged.

### Test fails on develop

```text
time_lines: object 'time_lines' does not exist
wall_hit_valid: object 'wall_hit_valid' does not exist
lwall_from_vmec: object 'lwall_from_vmec' does not exist
```

Terminal position can also describe the intersection while time and dynamics remain at the callback endpoint.

### Test passes after fix

```text
$ ./build_all -o clean_release -j 24 LIBSTELL BEAMS3D
BUILD COMPLETE

$ make test_beams3d_loss MPI_RUN_OPTS_SM='--ntasks=1'
STATUS: PASS

$ make test_beams3d_loss MPI_RUN_OPTS_SM='--ntasks=2'
STATUS: PASS

$ ./check_event_modes.py --case all --executable ../../BEAMS3D/Release/xbeams3d
BEAMS3D event-mode checks: PASS
```

The focused modes cover LSODE, RKH68, hit-only storage, a field-valid full-orbit hit, and an external-wall hit outside field support. Face counts and wall power remain first-hit quantities.


lazersos:
This changes a significnat portion of the codebase.  Can you explain why this is needed. The code correctly outputs the location where a particle hits a wall element. Knowing the exact values at the wall hit can be determined a-posterior for the most part. I suspect this would significantly slow the code down. I'd suggest just doing this as a post-processing in python if you need such information.

krystophny:
Fair question, and I should have led with this. The problem is that the quantities needed at the hit never reach the HDF5 file, so post-processing has nothing to reconstruct from. At a GC wall hit on develop (`out_beams3d_gc.f90`):

- `collide()` finds the intersection parametrically, but the chord fraction (`tmin` in `collide_double`) is discarded; only the position triplet is written back (lines 134-141).
- The terminal saved row is therefore mixed: R/PHI/Z hold the intersection, while vll and moment hold the end of the detection substep (the integrator has already stepped past the wall), and S/U/B hold values at that beyond-wall endpoint. With `-plasma` that endpoint lies outside the LCFS, where the splines have no physical support.
- `t_end` in the file is `t_last`, the substep end time (`beams3d_follow_gc.f90:156,211,299`), so the event time is known only to within one substep, and there is no per-save time dataset to do better.
- Neither chord endpoint is stored: the substep start exists only in `xlast` in memory, and the substep end position is overwritten by the intersection. Without an endpoint the fraction cannot be recomputed offline.
- The deposited GC wall power calls `fgc_eom` with `q2`, whose fourth component (vll) is never assigned (lines 136-150), so `wall_load` currently uses an uninitialized parallel velocity. `vperp` in the same expression is only set while the endpoint is inside the field box (line 95) and is stale otherwise.

As far as I can tell, reconstructing the hit time, vll/mu at the hit, or the deposited energy offline needs information the file does not contain, but I may be missing something: how would you reconstruct these from the current output?

On speed: the per-substep additions are a handful of scalar stores (`time_lines` plus sentinels); the field evaluation at the intersection runs once per lost particle, not per step. The integrator and stepping are untouched, and `test_beams3d_loss` passes at one and two ranks as in the PR description. I can attach wall-clock comparisons if useful.

On size: agreed, this PR carries more than the core fix. A minimal sub-step would be:

1. an optional `hit_fraction` output on `collide()` (backward compatible, about 10 lines, already in this branch);
2. per-particle face index, hit fraction, event time `t_prev + f*dt`, and vll/mu interpolated to the hit, used in `wall_load`/`wall_shine` (this also removes the uninitialized `q2(4)` path);
3. the corresponding nparticles-sized output datasets.

That is roughly 60-80 lines in `wall_mod` and the two callbacks: no per-save timestamps, no extra field diagnostics, no model or boundary-source flags. If that scope works for you I will open it against develop and keep the rest for separate discussion.

PR #482, opened 2026-07-12
Bugfix/linearto3 dindexing

lazersos:
This pull request improve the logic of the linear to 3D indexing we do.  In the previous version we had
```fortran
i = MOD(s-1,nr)+1
j = MOD(s-1,nr*nphi)
j = FLOOR(REAL(j) / REAL(nr))+1
k = CEILING(REAL(s) / REAL(nr*nphi))
```
which is now changed to the following per the testing of @krystophny 
```fortran
i = MOD(s-1,nr)+1
j = MOD(s-1,nr*nphi)/nr+1
k = (s-1)/(nr*nphi)+1
```
Both the regression tests for BEAMS3D and FIELDLINES pass with this change.  There may be similar improvements elsewhere in the code but I did not find them (maybe in STELLOPT somehwere). @krystophny has exenstively tested this and shown that the change fixes issues with large grids. @luisahuebner and @lvanham can you confirm that this change does not affect the current status of your results?

### DIFF
--- a/.github/workflows/build_use_docker.yml
+++ b/.github/workflows/build_use_docker.yml
@@ -20,5 +20,5 @@ jobs:
           ./build_all -o clean_debug -j 2
       - name: test
         run: |
-          make -C BENCHMARKS test_indexing
+          python3 -m unittest discover -s BENCHMARKS/tests -v
           mpiexec ./bin/xstelloptv2 test -h
--- a/.github/workflows/build_use_docker.yml
+++ b/.github/workflows/build_use_docker.yml
@@ -17,7 +17,7 @@ jobs:
       - name: compile
         run: |
           cd ${STELLOPT_PATH}
-          ./build_all -o clean_debug -j 2
+          ./build_all -o clean_debug -j 1
       - name: test
         run: |
           python3 -m unittest discover -s BENCHMARKS/tests -v
--- a/.github/workflows/build_use_docker.yml
+++ b/.github/workflows/build_use_docker.yml
@@ -20,4 +20,5 @@ jobs:
           ./build_all -o clean_debug -j 2
       - name: test
         run: |
+          make -C BENCHMARKS test_indexing
           mpiexec ./bin/xstelloptv2 test -h
--- a/BEAMS3D/Sources/beams3d_beam_density.f90
+++ b/BEAMS3D/Sources/beams3d_beam_density.f90
@@ -24,7 +24,7 @@
 !-----------------------------------------------------------------------
       IMPLICIT NONE
       LOGICAL :: lline_in_box
-      INTEGER :: ier, l, nl, i, j, k, m, s
+      INTEGER :: ier, l, nl, i, j, k, m, s, nrm1, nphim1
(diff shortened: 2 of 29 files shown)


krystophny (changes requested):
The integer decode is correct and matches what I validated for #479, but the TORLINES hunk has a typo that breaks compilation, so requesting changes for that.

## Decode equivalence

Exhaustive enumeration of old vs. new `(i,j,k)` over the full linear range, gfortran -O2, default 32-bit REAL as in the production builds:

| grid | N | old vs. new mismatches | first mismatch |
| --- | --- | --- | --- |
| 256 x 128 x 256 | 8388608 | 0 | none |
| 384 x 192 x 384 | 28311552 | 156 | s = 16809985 |
| 5000 x 5000 x 2 | 50000000 | 3291 | s = 16780000 |

Mismatches start just above 2^24 = 16777216, where `REAL(s)` loses integer resolution; the 5000 x 5000 case also corrupts `j` because `nr*nphi` itself exceeds 2^24. The new decode satisfies `i + (j-1)*nr + (k-1)*nr*nphi == s` at every point tested, and it is the same arithmetic I ran for #479, where the 1-, 2-, and 48-rank VMEC grids came out bitwise identical. The math is good.

## TORLINES typo (breaks compile)

`torlines_init_external.f90` lines 136, 192, 225: `v = (i-1)/(nrrho*nu)+1` should read `nrho`. The file has `IMPLICIT NONE` and `nrrho` is declared nowhere, so TORLINES no longer compiles on this branch. CI stays green because `build_all` does not propagate per-code make failures and the test step only runs xstelloptv2.

## Remaining sites with the same pattern

Could be fixed here or in a follow-up:

- `FIELDLINES/Sources/fieldlines_gridgen.f90:224`
- `STELLOPTV2/Sources/General/stellopt_write_mgrid.f90:211-212` (large mgrids can pass 2^24 points)
- `STELLOPTV2/Sources/General/stellopt_compute_bnormal.f90:146,285`
- `LIBSTELL/Sources/Modules/wall_mod.f90:1539-1540` (block decode; block counts stay small today)

The `FLOOR(REAL(n)/REAL(nprocs))` chunk computations elsewhere share the pattern but only fail past 2^24 items, which nothing reaches right now.

I can add a small Python regression to BENCHMARKS that enumerates the decode across the 2^24 boundary so this cannot regress silently; say the word and I push it onto this branch.

krystophny:
@lazersos you may have noticed that Fable is acting on my behalf these days. I'm burning the last tokens for useful things. I will go over things manually then later during the week.

lazersos:
@krystophny, yes please push the test to this branch.

krystophny:
Added the requested linear-to-3D regression in b88c5a2c (final head 603265f7). It exhaustively round-trips indices across the 2^24 single-precision boundary, pins the known old REAL-based failure at s=16809985, and is wired into BENCHMARKS and CI. Validation: 2 local tests passed; both GitHub Actions builds pass on the final head.

krystophny (approved):

PR #483, opened 2026-07-13
Feature/cn normalized

ajchcoelho:
This adds the possibility to read a normalized pinch from the sources file and add it to the particle solver in THRIFT. This external pinch is denormalized internally using the gyro-Bohm scaling.


lazersos (approved):
Looks good to me.

PR #488, opened 2026-07-24
Addition of the Henneberg boundary representation into STELLOPT

lazersos:
This pull request adds to STELLOPT the ability to optimize configurations in the Henneberg space of equilibrium representations. It also provides python interface to the forward and back tranformations from VMEC RBC/ZBS.


ajchcoelho:
Looks good to me.

PR #489, opened 2026-07-27
Bugfix/fieldlines bcyl

luisahuebner:
Fixes an error in a fieldlines coordinate transform subroutine.


lazersos:
@lvanham can you look through this and verify that you're OK with our inclusion of your modifications to MUMAT?  If not we can rework the change becasue really the bugfix is limited to one routine related to FIELDLINES and MUMAT.

luisahuebner:
I just noticed that this merge would overwrite some of the recent changes to the linear to 3D indexing. Instead of the new:
i = MOD(s-1,nr)+1
j = MOD(s-1,nr*nphi)/nr+1
k = (s-1)/(nr*nphi)+1
this merge would change it to:
i = MOD(s-1,nr)+1
j = FLOOR(REAL(MOD(s-1,nr*nphi))/REAL(nr))+1
k = CEILING(REAL(s)/REAL(nr*nphi))

I have a test run in the queue to verify that the new indexing doesn't change the results. If that goes well I will modify fieldlines_init_mumat to use the new indexing.

lvanham (approved):

PR #490, opened 2026-07-28
Add check for wall faces with index == 0 and convert to 1-based indexing

kudav:
Straightforward implementation of a guard against accidentally using 0-based walls. The handling is transparently printed to the output log. I consider the chance of a 1-based wall face file containing a 0 face negligible.


lazersos (approved):
Looks good.

PR #491, opened 2026-07-31
Feature/kinetic profile files

kudav:
ASTRA writes pressure, iota and current profiles to file each transport step and drives VMEC with them.  
In this PR, vmec_input.f gains the PMASS_FILE / PIOTA_FILE / PCURR_FILE namelist entries and profile_functions.f a 'file' branch for each of the three profiles, so VMEC reads the profile from disk.

A small bugfix was needed to restore compilation, after #490 put a declaration inside an executable IF block.

### DIFF
--- a/LIBSTELL/Sources/Miscel/profile_functions.f
+++ b/LIBSTELL/Sources/Miscel/profile_functions.f
@@ -54,13 +54,14 @@
 !  using a 10-point Gauss-Legendre quadrature.
       USE stel_kinds
       USE stel_constants, ONLY: zero, one, pi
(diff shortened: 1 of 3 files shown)


lazersos:
Can't do it this way. We can't add 'features' to the INDATA namelist without going through the ORNL branch and making sure this feature is properly documented. I would suggest if you need this type of interfacing that you fork VMEC from ORNL and then add that to your fork.  Better would be to modify ASTRA linking to LIBSTELL and setting the AC/AI/AM_AUX_S/F arrays before writing....or just use THRIFT, which already does what ASTRA does and more. 😈

PR #494, opened 2026-08-07
VMEC_UTILS: improve inverse lookup stability and flexibility

kudav:
# VMEC_UTILS: improve inverse lookup stability

## The situation
The `(R,Z) -> (s,u)` inverse map in `cyl2flx`/`newt2d` 
- returned **whatever the last restart produced** rather than the best one
- could hand back a coordinate that was not the one it measured its residual at.
- had its tolerance, damping, iteration cap and restart count hard-wired

## What's now implemented
- `cyl2flx` retains the lowest-residual restart instead of the last one,
  which is also what fixes the angle-jog defect below
- `GetBcyl_WOUT` accepts a solution already inside `fmin_acceptable`, as
  `GetAcyl_WOUT`, `GetJcyl_WOUT` and `MSE_pitch_VMEC` always have
- off-by-one fix in the `GetAcyl_WOUT` radial index
- optional boundary re-seeding for points the map places just outside a
  sharply-shaped LCFS
- the solver controls become tunable module variables, settable from
  `&beams3d_input`, **with defaults that reproduce today's behaviour**
- `-DDEBUG_CYL2FLX` diagnostics plus a pySTEL reader/plotter
- BEAMS3D warns when the map fails *near the plasma*
- BEAMS3D resets the `sflx`/`uflx` seed on `ier == -3` as well as `-1` —
  the fix raised in the #493 review and not picked up when that PR landed.
  With the warm start `sflx`/`uflx` seed the next grid point, and on
  `ier == -3` `GetBcyl` returns no field and leaves them wherever the
  diverged Newton stopped, so the worst seeds were exactly the ones that
  propagated. Two lines, no change to the seeding scheme itself.
- a plasma-only FIELDLINES regression test, which nothing covered before

Rebased onto develop after #478, #480, #488 and #493. Note that #493
landed with the hoisted warm start, so `beams3d_init_vmec` still seeds
each lookup from the previous grid point; this branch does not reopen
that, it only adds the `ier == -3` seed reset asked for in the #493
review.

## Nothing is lost, and the cells gained are inside the plasma

Resolved-cell counts, `S_ARR <= 1`, 4 ranks, remeasured against develop
`fcb85068`:

| case | grid | develop | this branch | delta |
|---|---|---|---|---|
| ORBITS | 128x32x128 | 404 480 | 404 608 | +128 |
| W7X | 128x72x128 | 205 350 | 205 416 | +66 |
| AUG | 101x16x101 | 55 952 | 55 968 | +16 |

No case loses a cell. The branch's three counts are bit-for-bit the ones
measured before #478/#493 landed; develop's have slipped a little since,
so the deltas above are develop moving, not this branch. I have not
attributed the slip to a specific commit in that range.

Those deltas are not cells at the edge. Counting cells that lie **inside**
the forward-mapped LCFS and are still flagged `S_ARR = 4` — a point in the
plasma that BEAMS3D then treats as vacuum, with no field and no flux
label — gives, on W7X at 4 ranks:

| | develop | this branch |
|---|---|---|
| W7X, full boundary | 66 | **0** |
| W7X, boundary truncated to `m <= 5` | 85 | **0** |

FIGURE-4-w7x_interior_failures

*W7X, full boundary. Every cell that lies inside the LCFS and is still
flagged `S_ARR = 4`, pooled over all 72 toroidal planes onto the shared
(R,Z) grid — the plasma outline is therefore the envelope swept by the
LCFS over the field period, with the bean (phi = 0, solid) and triangle
(phi = 36 degrees, dashed) cross sections marked. develop leaves 66 such
cells; this branch leaves none.*

FIGURE-2-rz_AUG_delta

*AUG, phi = 0, 4 ranks, `-plasma -depo`, against develop
`fcb85068`. Left: `B_phi(new) - B_phi(old)`, symmetric log scale with the
linear band at 1e-3 T. Right: cells that cross the `S_ARR = 4` sentinel.
Black is the wout LCFS.*

Two things to read off it. First, the change is bimodal and nothing in
between: of the 36 841 cells in the volume whose `B_phi` differs at all,
36 297 differ by less than 1e-3 T — in fact by ~1e-13 T, last-bit, from
the retain-best path selecting a different restart — and the remaining
544 are the boundary cells where the map went from returning nothing to
returning a field. The plasma interior does not move.

Second, the sentinel crossings are not symmetric noise: 336 cells in the
volume leave `S = 4` and 144 enter it, and the ones that leave are real.
Of the 21 in this plane, 20 solve to 1.15 < s < 1.99 — far-SOL points
that were being discarded. The twenty-first sits at R = 1.4023,
Z = +0.5707, **inside** the LCFS, where develop returns the sentinel and
this branch solves it to s = 0.8697.)

## The angle-jog defect (@pponsv)

@pponsv spotted that `cyl2flx` jogs the poloidal angle *after* `newt2d`
returns, so when the restart loop runs out the returned angle is the solved
angle plus one uncompensated jog — while `fmin` still describes the
pre-jog coordinate. Confirmed, and it is worse than a cosmetic mismatch.

The check needs no reference data: `cyl2flx` reports both `fmin` and
`c_flx`, and both come from the same forward map, so re-evaluating
`flx2cyl` at the returned `c_flx` has to reproduce `fmin` to roundoff.
On develop it does not. AUG, 60x3x60 over phi = 0, 1/4,
1/2 field period:

| | develop | this branch |
|---|---|---|
| points with a claimed solution | 8319 | 8451 |
| residual at returned coord > 4x reported `fmin` | **465** | **0** |
| of those, interior (`s <= 1`) | 3 | 0 |

The mechanism is exact, not inferred. At the worst offender, undoing one
`twopi/20` jog reproduces the reported `fmin` bit-for-bit:

```
u - twopi/20 : residual 1.9490E-13   RZ miss 0.000001 m
u  (as ret.) : residual 3.6854E-02   RZ miss 0.235582 m
reported fmin was     : 1.9490E-13
```

It reaches the public API, because `fmin = 1.9e-13` passes the documented
`fmin_acceptable = 1e-8` gate and callers accept the point.
`GetBcyl_WOUT(R=1.200995, phi=0, Z=-0.252048)`:

| | develop | this branch |
|---|---|---|
| `info` | -1 | 0 |
| returned `(s, u)` | 0.963468, **3.668087** | 0.963468, **3.353931** |
| forward map of that `(s,u)` lands at | (1.2616, -0.4797) | (1.2010, -0.2520) |
| off by | **0.236 m** | 0.000 m |
| `B_phi` | -3.234 T | -3.567 T |

That is **B_phi wrong by 9.3%** at an interior point, `s = 0.963`. NCSX
shows the same signature (69 inconsistent points, worst 0.107 m); W7X
shows none.

Retaining the best restart fixes it structurally rather than by
re-ordering the jog: `xc_best` is snapshotted before any jog and restored
after the loop, so the returned coordinate is always the one `fmin` was
measured at. @pponsv's proposal to move the jog ahead of the `newt2d` call
reaches the same place; this branch already contains an equivalent fix.

The same check then found one more instance of the defect **in this
branch's own boundary recovery**: when re-seeding failed it snapped the
point onto the `s=1` node and reported `info=0` with the *previous*
coordinate's residual — `fmin ~ 1e-30` for a point 6.9 mm out. That
fallback is now deleted; a re-seeded try competes on residual like any
other, and if it does not come inside, the point stays outside.

## cyl2flx defaults at their usual values

Summary of tolerance values:

| knob | first proposed | now | why |
|---|---|---|---|
| `CYL2FLX_FTOL` | 1e-12 | 1e-16 | 1e-16 *is* reachable — 99.7% of ORBITS points hit it, at 2.5% of the iteration budget. Costs 0.3–5%. |
| `CYL2FLX_NRESTART` | 8 | 4 | 8 restarts recovered 0 cells and cost ~43 s of the W7X grid build. |
| `CYL2FLX_LBAIL_OUTSIDE` | never bail | bail | Not bailing recovered 0 cells on AUG, NCSX and W7X, and cost ~36 s on W7X. |

W7X grid build, 128x72x128, 4 ranks, all four configurations remeasured
against develop `fcb85068`. Five repetitions cycled config-by-config, so
drift in machine load spreads over all four rather than landing on
whichever ran last; median and range over the repetitions:

FIGURE-1-w7x_cost

| configuration | time | resolved |
|---|---|---|
| develop | 26.8 s (21.9–27.4) | 205 350 |
| 8 restarts, no bail | 113.7 s (109.6–119.2) | 205 416 |
| 4 restarts, no bail | 70.9 s (70.2–73.3) | 205 416 |
| **4 restarts, bail (this branch)** | **35.4 s (34.7–36.1)** | **205 416** |

Eight restarts recover nothing over four, and never bailing recovers
nothing over bailing — all three branch configurations land on the same
205 416, while develop sits 66 below. (`4 restarts, no bail` differs from
the other two by two cells in the `1 < s < 4` band, none in `s <= 1`.) So
the two expensive knobs buy nothing on this equilibrium, which is why they
default to the cheap setting.

That is 1.32x develop's grid-build time on this equilibrium, for 66
in-plasma cells develop leaves unresolved plus the 465 AUG points above.
The gain is correctness, not coverage. (develop's 21.9 s low is its first
repetition and sits well below its other four, which cluster 26.7–27.4;
the per-repetition ratio over the remaining four is 1.29–1.34x.)

`CYL2FLX_LBNDRY` stays on by default. It recovers nothing here (W7X
205 416 either way, cost inside run-to-run noise) because it is aimed at
equilibria whose mode count is low relative to their spectral
condensation, where the boundary is rounded and tips fall outside. None
of AUG (mnmax 13), NCSX (137) or W7X (263) is mode-starved, so this does
not disprove the mechanism — it shows it does not trigger here.

The warning below is how you find out when the mechanism does trigger.

## The warning

BEAMS3D now reports grid cells where the map failed near the plasma. It
excludes failures at `s >= 1.5`: those are exterior points with no solution,
`S=4` is correct for them, and counting them would fire on every healthy run.

On AUG it reports **0.67% of the grid (1088 cells)**, remeasured on the
current base at 4 ranks with `-plasma -depo`. Before this branch that was
silent.


pySTEL/libstell/cyl2flx_diag.py
```diff
@@ -0,0 +1,447 @@
...
+	Z = v.sfunct(theta, phi, v.zmns, v.xm, v.xn)
+	if getattr(v, 'iasym', 0) == 1:
+		R = R + v.sfunct(theta, phi, v.rmns, v.xm, v.xn)
+		Z = Z + v.cfunct(theta, phi, v.zmnc, v.xm, v.xn)
+	Rb = np.asarray(R[s_index, :, 0])
+	Zb = np.asarray(Z[s_index, :, 0])
+	return Rb, Zb, int(v.nfp)
+
+
+if __name__ == '__main__':
```
lazersos: Class file mains should always look like:
```python
# Main routine
if __name__=="__main__":
	import sys
	sys.exit(0)
```

lazersos (dismissed):
Slight change in the proposed class file, otherwise looks good.

krystophny (approved):

PR #495, opened 2026-08-07
Bugfix/beams3d fo fix

kudav:
Full orbit is currently broken in three independent ways, and nothing in the test suite would notice any of them. This branch fixes all three and adds benchmarks that fail on `develop`.

1. The collision operator lets markers gain energy without bound.
2. Two of the three integrator branches — `NAG` and `RKH68` — cannot produce a physical trajectory at all.
3. The one accuracy knob the fixed-step integrator has is a hardcoded constant.

Particularly using the NUBEAM collision operator, the markers received large kicks in energy due to three defects:
fig1_collisional_runaway

Fixed the RKH68 integrator for full orbit - was not reachable before. For full orbit, this integrator performs better at the same cost as LSODE:
fig3_integrator
To tune RKH68 tolerance, the NSUB_FULLORBIT namelist parameter was introduced, exposing NSUB in beams3d_calc_dt.

Note that running LSODE at default 1e-8 tolerance is not advised, as the energy drift is too large, 1e-10 or less is necessary:
fig2_collisionless_drift

With these fixes, collisional runs work again, which fail currently on develop, and the benchmark from the 2023 paper on full orbit is reproduced:
fig4_banana


lazersos (approved):
Looks good to me.

PR #500, opened 2026-08-18
Call penta library from pySTEL transport solver

ajchcoelho:
We can now run pySTEL transport solver with the neoclassical fluxes computed from `PENTA`. This is done via the newly created `libpenta` library, which was written in the same philosophy as libstell. Some changes to `STELLOPT` makefiles had to be done.


lazersos (approved):
Looks good to me.

PR #502, opened 2026-08-19
Deallocate the arrays neo_dealloc tests, and the LASYM ones

rogeriojorge:
Two guards in `NEO/Sources/neo_dealloc.f90` free a different array than the one they test:

```fortran
IF (ALLOCATED(pixm)) DEALLOCATE(pixn)
IF (ALLOCATED(i_m)) DEALLOCATE(i_n)
```

`pixm` and `i_m` are therefore never released, and `pixn`/`i_n` are freed on a condition that does not describe them — harmless while the pairs are allocated together, but it is not what the lines intend.

Separately, the four asymmetric spectra that `READ_BOOZ_IN` allocates under LASYM (`rmns`, `zmnc`, `lmnc`, `bmns`) are never deallocated, while their cosine partners are.

Both leak once per `neo_dealloc` call, which adds up inside a STELLOPT optimization loop that reinitializes NEO each iteration.


lazersos (approved):
Another good catch!

